PR #5, opened 2018-03-15
Create new BlockManager architectural component

aludvik:
Signed-off-by: Adam Ludvik <ludvik@bitwise.io>

### DIFF
--- a/text/0000-block-manager.md
+++ b/text/0000-block-manager.md
@@ -1,0 +1,367 @@
+- Feature Name: `block-manager`
+- Start Date: 2018-03-13
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC describes new architectural components of the validator that handle
+the management of blocks correctly and help remove a known race condition that
+can cause network nodes to fork.
+
+# Motivation
+[motivation]: #motivation
+
+The RFC aims to improve Sawtooth in two ways:
+
+1. By removing a race-condition  which can cause parent blocks required for
+   validation of some block to be dropped while the block is transferred from
+   the completer to the chain controller.
+2. Providing additional guarantees about the blocks being processed.
+3. Simplifying and generalizing the procedure for persisting blocks.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC aims to improve the handling of blocks by introducing some new classes
+and interfaces. The requirements for the design presented in this RFC are that
+it:
+
+- Shall store blocks for later
+- Shall ensure integrity of blocks stored
+- Shall provide a mechanism for preventing required blocks from being dropped
+- Shall provide a mechanism for persisting to a store
+
+The constraints for the design presented in this RFC are that:
+
+- The API must work for both forking vs non-forking consensus mechanisms
+- The API defined must be reusable, meaning it provides generic methods that
+  make few assumptions about how the API will be used
+
+The following new concepts are used by this design and are defined here:
+
+For a collection of blocks, __integrity__ is defined to mean that every block
+in the collection is accompanied by its parent.
+
+A __branch__ is a collection of blocks such that:
+
+1. The blocks are sorted in order of increasing block number
+2. The branch has integrity
+
+A __block tree__ is a collection of branches with a single common ancestor or
+__root__.
+
+The following is a visualization of an example block tree containing blocks
+{0, A, B, C, ..., H}.
+
+```
+0 (root)
+|
+|\
+A |
+| E
+|\
+B |
+| F
+| |\
+C | |
+| G |
+|   H
+D
+```
+
+In the above example, the block tree contains multiple branches. They are (0,
+A, B, C, D), (0, E), (0, A, F, G), and (0, A, F, H).
+
+In order to satisfy the requirements and constraints of this RFC, the
+__block manager__ class and __block store__ interface were created.
+
+A __block store__ is an interface which supports storing and retrieving blocks.
+It can be backed by any type of storage, for example disk or network. Because a
+blockchain's length grows indefinitely, part of it eventually has to be
+archived in some way. The __block store__ interface is intended to help solve
+the problem of persisting parts of the __block tree__ in a way that does not
+make assumptions about how it is persisted or what parts of the tree are
+persisted.
+
+The __block manager__ is a collection of __block tree__ s (though it usually
+contains 1) that also supports methods for:
+
+- Ensuring that blocks depended on by external objects are not removed.
+- Persisting blocks to a __block store__.
+
+In addition to the __block manager__ class and __block store__ interface, a
+__commit store__ class was created The commit store is a __block store__
+implementation that can be added to the block tree for persisting committed
+blocks to disk. It also implements an LRU cache which holds some number of most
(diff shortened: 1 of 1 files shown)


text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+The following is a listing of the __block manager__ methods and their function:
+
+`put(branch: List<Block>)`
+
+Atomically adds the given blocks to the tree.
+
+The blocks passed must form a branch meaning they satisfy the following
+conditions:
+
+1. The first block's parent is present
```
boydjohnson: Does present mean that the parent block is already in the block manager? It is ambiguous as to whether is is that or "present in the List<Block>".

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+completed a block B and would like to pass it to the chain controller. If the
+completer were to just pass the chain controller B through a queue or shared
+memory, the chain controller could decide to `unref_block` A before it receives
+B, causing B to have a missing predecessor when it arrives at the chain
+controller. Instead, the completer first places a `ref_block` on A, causing and
+then it passes B to the chain controller. If the chain controller decides to
+`unref_block` A now, the block manage will ensure it is not actually dropped,
+since there is still an open hold on the block. Finally, when the chain
+controller takes ownership of B, it can send a signal back to the completer
+that it has been received and the completer can release its hold.
```
boydjohnson: "completer can release its hold": Does the completer have a hold on the block or is it the Block Manager?

text/0000-block-manager.md
dcmiddle: I read that to mean that the parent of the list[0] element is already present in tree managed by the block manager.

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+to represent a block that is in a store and not in the __main cache__ and
+consists of:
+
+- The block id of the block that is in the store
+- A list of names of stores that the block is in
+- A reference count of blocks that depend on the anchor but are not in a store
+  the block is in
+
+A block in a store and not the __main cache__ has an implicit reference count
+of 1 if it is not the tip of the branch in the store. If it is the tip of the
```
dcmiddle: of 1 or of _at least_ 1? (viz holds placed by other objects)

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The following provides more detail on how the API present above is implemented.
+
+## Internal Representation
+
+Internally, the block manager consists of:
+
+- A __main cache__ which, at a minimum, contains all blocks that have not been
```
dcmiddle: is main cache a block store?

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+
+The __block manager__ is a collection of __block tree__ s (though it usually
+contains 1) that also supports methods for:
+
+- Ensuring that blocks depended on by external objects are not removed.
+- Persisting blocks to a __block store__.
+
+In addition to the __block manager__ class and __block store__ interface, a
+__commit store__ class was created The commit store is a __block store__
+implementation that can be added to the block tree for persisting committed
```
dcmiddle: What does it mean to say that the commit store can be added to the block tree? The block tree class has a member which which implements store? Does a tree have to be instantiated with a store implementation? I'm generally unclear on the relationship between block stores and trees. Maybe a fuller description of the block tree class would help.

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+persisted.
+
+The __block manager__ is a collection of __block tree__ s (though it usually
+contains 1) that also supports methods for:
+
+- Ensuring that blocks depended on by external objects are not removed.
+- Persisting blocks to a __block store__.
+
+In addition to the __block manager__ class and __block store__ interface, a
+__commit store__ class was created The commit store is a __block store__
```
dcmiddle: Is the commit store a member of the block manager?

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+    removed from the tree.
+
+    Raises an exception if the reference count falls below 0, which indicates
+    a bug in the application code.
+
+The following example illustrates how the `ref_block` operation is used to
+prevent blocks from being dropped when they are transferred from the completer
+to the chain controller.
+
+Assume that the chain controller has a block A and that the completer has just
```
dcmiddle: Do the chain controller and completer both hold references to the same block manager or does the chain controller own the block manager? I assume the former. In which case the controller and the completer must both signal the block manager before / after certain operations.

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+completed a block B and would like to pass it to the chain controller. If the
+completer were to just pass the chain controller B through a queue or shared
+memory, the chain controller could decide to `unref_block` A before it receives
+B, causing B to have a missing predecessor when it arrives at the chain
+controller. Instead, the completer first places a `ref_block` on A, causing and
```
dcmiddle: To clarify regarding above comment about ownership vs. signaling of the block manager...
"Instead, the completer first calls the __block manager__ to place a `ref_block` on A ~causing~ and"

text/0000-block-manager.md
dcmiddle: And what does it mean for the chain controller to take ownership of B? Would the completer actually be passing a reference to the controller for an object managed by the block manager?

text/0000-block-manager.md
```diff
@@ -0,0 +1,367 @@
...
+In order to avoid keeping a reference count and copy of every persisted block
+in the main cache, __anchors__ are placed in the main cache to indicate that
+the rest of the branch has been persisted to some store. An __anchor__ is used
+to represent a block that is in a store and not in the __main cache__ and
+consists of:
+
+- The block id of the block that is in the store
+- A list of names of stores that the block is in
+- A reference count of blocks that depend on the anchor but are not in a store
+  the block is in
```
dcmiddle: This seems confusing to have reference counting on the anchor and on the block. Would it make sense to just have reference counting managed by the tree? I'm assuming all blocks are referenced in the tree(s)?

dcmiddle (changes requested):
Sorry for the late review. Appreciate responses when you are able.

text/0000-block-manager.md
boydjohnson: Hi, @dcmiddle . Since I am implementing this right now, I thought I should respond. The main cache is in-memory and is not a block store. It is a logical space. Right now in my implementation it is a HashMap of block ids to Block - reference count tuples, wrapped in a lock. Specifically Rc<RwLock<HashMap<String, (Rc<Block>, i64, i64)>>>.

text/0000-block-manager.md
aludvik: The block manager has a map of store names to stores. The commit store would be one such member of that map.

text/0000-block-manager.md
aludvik: A block store is just a place to persist blocks, specifically a single branch of blocks. Multiple stores can be added to the manager and branches can be persisted to multiple stores. So, for example, you could have a faster local disk store and slower, perhaps network backed, store for archiving very old blocks.

PR #6, opened 2018-03-22
Establish Sawtooth Governance Model

aludvik:
This RFC proposes an explicit governance structure for the Hyperledger Sawtooth project and establishes the initial teams and members that makeup this structure. It also defines the meaning of committer and maintainer for individual repositories.

### DIFF
--- a/text/0000-sawtooth-governance.md
+++ b/text/0000-sawtooth-governance.md
@@ -1,0 +1,728 @@
+- Feature Name: not applicable
+- Start Date: 2015-02-27
+- RFC PR: [rust-lang/rfcs#1068](https://github.com/rust-lang/rfcs/pull/1068)
+- Rust Issue: N/A
+
+# Summary
+
+This RFC proposes to expand, and make more explicit, Rust's governance
+structure. It seeks to supplement today's core team with several
+*subteams* that are more narrowly focused on specific areas of
+interest.
+
+*Thanks to Nick Cameron, Manish Goregaokar, Yehuda Katz, Niko Matsakis and Dave
+ Herman for many suggestions and discussions along the way.*
+
+# Motivation
+
+Rust's governance has evolved over time, perhaps most dramatically
+with the introduction of the RFC system -- which has itself been
+tweaked many times. RFCs have been a major boon for improving design
+quality and fostering deep, productive discussion. It's something we
+all take pride in.
+
+That said, as Rust has matured, a few growing pains have emerged.
+
+We'll start with a brief review of today's governance and process,
+then discuss what needs to be improved.
+
+## Background: today's governance structure
+
+Rust is governed by a
+[core team](https://github.com/rust-lang/rust-wiki-backup/blob/master/Note-core-team.md),
+which is ultimately responsible for all decision-making in the
+project. Specifically, the core team:
+
+* Sets the overall direction and vision for the project;
+* Sets the priorities and release schedule;
+* Makes final decisions on RFCs.
+
+The core team currently has 8 members, including some people working
+full-time on Rust, some volunteers, and some production users.
+
+Most technical decisions are decided through the
+[RFC process](https://github.com/rust-lang/rfcs#what-the-process-is).
+RFCs are submitted for essentially all changes to the language,
+most changes to the standard library, and
+[a few other topics](https://github.com/rust-lang/rfcs#when-you-need-to-follow-this-process).
+RFCs are either closed immediately (if they are clearly not viable),
+or else assigned a *shepherd* who is responsible for keeping the
+discussion moving and ensuring all concerns are responded to.
+
+The final decision to accept or reject an RFC is made by the core
+team. In many cases this decision follows after many rounds of
+consensus-building among all stakeholders for the RFC. In the end,
+though, most decisions are about weighting various tradeoffs, and the
+job of the core team is to make the final decision about such
+weightings in light of the overall direction of the language.
+
+## What needs improvement
+
+At a high level, we need to improve:
+
+* Process scalability.
+* Stakeholder involvement.
+* Clarity/transparency.
+* Moderation processes.
+
+Below, each of these bullets is expanded into a more detailed analysis
+of the problems. These are the problems this RFC is trying to
+solve. The "Detailed Design" section then gives the actual proposal.
+
+### Scalability: RFC process
+
+In some ways, the RFC process is a victim of its own success: as the
+volume and depth of RFCs has increased, it's harder for the entire
+core team to stay educated and involved in every RFC. The
+[shepherding process](https://github.com/rust-lang/rfcs#the-role-of-the-shepherd)
+has helped make sure that RFCs don't fall through the cracks, but even
+there it's been hard for the relatively small number of shepherds to
+keep up (on top of the other work that they do).
+
+Part of the problem, of course, is due to the current push toward 1.0,
+which has both increased RFC volume and takes up a great deal of
+attention from the core team. But after 1.0 is released, the community
+is likely to grow significantly, and feature requests will only
+increase.
+
+Growing the core team over time has helped, but there's a practical
+limit to the number of people who are jointly making decisions and
+setting direction.
+
+A distinct problem in the other direction has also emerged recently: we've
+slowly been requiring RFCs for increasingly minor changes. While it's important
+that user-facing changes and commitments be vetted, the process has started to
+feel heavyweight (especially for newcomers), so a recalibration may be in order.
+
+We need a way to scale up the RFC process that:
+
+* Ensures each RFC is thoroughly reviewed by several people with
+  interest and expertise in the area, but with different perspectives
+  and concerns.
+
+* Ensures each RFC continues moving through the pipeline at a
+  reasonable pace.
+
+* Ensures that accepted RFCs are well-aligned with the values, goals,
+  and direction of the project, and with other RFCs (past, present,
+  and future).
+
+* Ensures that simple, uncontentious changes can be made quickly, without undue
+  process burden.
+
+### Scalability: areas of focus
+
+In addition, there are increasingly areas of important work that are
+only loosely connected with decisions in the core language or APIs:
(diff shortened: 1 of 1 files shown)


aludvik:
Initial Root Team membership will consist of (sorted by last name):
* Shawn Amundson
* Andi Gunderson
* Adam Ludvik
* Dan Middleton
* James Mitchell
* Peter Schwarz

aludvik:
This RFC has already undergone substantial internal discussion and is entering the Final Comment Period immediately. Members of the initial root team, please indicate your approval by adding a reaction to this comment.

vaporos:
I approve for FCP.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+
+A **maintainer** has permission to approve changes. All pull-requests must be
+approved by at least 2 maintainers before a committer may merge it.
+Maintainers are expected to:
+
+- Carefully review pull requests and leave thoughtful, constructive feedback
+- Ensure the component is up-to-date with patches and security updates
+- Participate in community discussions related to their component
+
+A committer can be promoted to a maintainer by unanimous vote by the existing
```
delventhalz: Does unanimous in this context mean everyone votes for or no one votes against? For the root team, this ambiguity is probably fine, but the list of maintainers might grow quite large, and it might no longer be practical to get an affirmative vote from everyone. So for maintainers in particular, I would suggest the "no one votes against" version, but I'd like to clarify what unanimous means either way.

aludvik: I think the implicit assumption here is that the maintainer list, which is on a per-component/per-repository basis, stays relatively small and is only comprised of the individuals with an expert knowledge and understanding of the code-base. If the maintainer list on a repo grows so large that it is impractical to get an affirmative vote from everyone, then I would interpret that as an indication that the group has grown too large. The intent is to make decisions by consensus and it is the responsibility of those in leadership positions to "drive consensus" by keeping people involved.

TomBarnes:
Where does the core subteam membership get documented?

ineffectualproperty:
I'd like to see a broader set on the 'Initial Root Team' beyond code contributors. I have been actively involved in the project since it's inception so would like to have a voice w.r.t. RFCs. I know @TomBarnes has been instrumental in compliance activities and Mic Bowman from the research perspective.

text/0000-sawtooth-governance.md
grkvlt: I'd agree with the suggestion by @delventhalz since this is how an ASF PMC also operates, as long as there are no negative votes and at least one binding positive vote by a PMC member, a proposal passes.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+some existing community members may feel offended by not being given a
+permission level that is consistent with their expectations or by not being
+placed on a team whose domain they feel responsible for.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The [Mozilla governance model](https://www.mozilla.org/en-US/about/governance/policies/module-ownership/ )
+was also considered, however a hierarchical model was focused on ensuring the
+project maintained a consistent vision was preferred.
```
grkvlt: Also http://www.apache.org/foundation/governance/pmcs.html which describes the ASF model.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+
+Members of the Sawtooth community are expected to abide by the [Hyperledger Code of Conduct][hyperledger-coc].
+Violations of this code of conduct by any community member may result in
+disciplinary action at the discretion of the root team (by unanimous vote,
+excluding the offending member if applicable).
+
+[hyperledger-coc]: https://wiki.hyperledger.org/community/hyperledger-project-code-of-conduct
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
grkvlt: This doesn't specify anything about e.g. release processes and strategies, or who can approve or veto a release of the software. I think this is the most important part, since the project is primarily about producing releases of the Sawtooth software.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+**Consensus** - Responsible for the supported consensus algorithms and
+the consensus interface
+
+**Seth** - Responsible for the Sawtooth integration with Ethereum
+
+**Community Outreach** - Responsible for writing and hosting documentation,
+managing the main Sawtooth website, promoting Sawtooth to the public, managing
+demos, and providing the community with training and support
+
+**Release Management** - Responsible for release related issues such as
```
grkvlt: I'm not convinced this should be a separate group? I think there are cross-cutting concerns like release and CI that should be handled by the top-level team, with sub-teams responsible for separate projects able to modify the processes within limits.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+tradeoffs against the project's priorities and values** (which the root team is
+responsible for setting, globally). In the end, these decisions are about how
+to weight tradeoffs. The decision should be communicated in these terms,
+pointing out the tradeoffs that were raised and explaining how they were
+weighted, and **never introducing new arguments**.
+
+### Initial Subteams
+
+The following is a proposed list of initial subteams which is subject to change
+at the discretion of the root team, once it has formed.
```
grkvlt: I would prefer if the sub-teams are focussed on software components that have their own separate existence, such as Seth and the Go SDK, where there is a definite releasable, versioned artifact they are responsible for. Other concerns like community outreach should be handled by the each team, taking their lead/direction from core.

text/0000-sawtooth-governance.md
```diff
@@ -0,0 +1,348 @@
...
+
+A **reviewer** has "read" permission, meaning they can review pull requests.
+
+A **committer** has "write" permission, meaning they can merge pull requests
+once they have been approved.
+
+Any community member can be promoted to a reviewer or committer by a
+maintainer. Reviewers and committers lose their status when it is removed by a
+maintainer. This is usually due to inactivity, but can also be used for
+moderation.
```
grkvlt: I'm not clear what exactly the point of the reviewer/committer/maintainer split is. It seems that a committer is basically a reviewer who is able to click the 'merge' button on github, but only once two maintainers have approved the PR. I think this is getting overly complicated, and really you only need a 'committer' role with write access privileges, plus the rule that merges can only occur once two committers have approved the change? Why would there be a special class of people trusted to review PRs but not trusted to approve or merge them? This sounds like they are just ordinary GitHub users, since the repos are public, so we don't need to specify their rights here.

text/0000-sawtooth-governance.md
aludvik: Thanks for the link

aludvik:
@TomBarnes @grkvlt The actual set of subteams and their membership rosters will be decided by the root team once it has formed.

text/0000-sawtooth-governance.md
delventhalz: I am still leaning towards a "no votes against" version of unanimous. If nothing else, maybe people are sick, or otherwise unavailable (though in that case, we probably need to spell out what a reasonable timeframe for a response on a vote is).

More importantly, however we're using the term, I think it is important to explicitly define unanimous.

text/0000-sawtooth-governance.md
aludvik: For the subteam breakdown, we tried to take inspiration from an existing model that seems to be working: https://www.rust-lang.org/en-US/team.html Their breakdown includes high-level, cross-cutting concerns such as a release team and a community team.

Re-reading this description again, I think it could be clarified that this subteam is responsible for the release management of the core platform and not necessarily separate components like the SDKs and Seth.

text/0000-sawtooth-governance.md
dcmiddle: I tend to agree with @grkvlt.
Maybe there's a mix here of what constraints the github permissions allow and what permissions and roles we think are necessary.
I'd like to invite any specific user to review a PR. This seems to be a limitation on github atm.
I'd like there to be a well defined limited set of people who can merge. Those individuals have the judgement to get two diverse approvals and know when it's ok to click merge. I don't care if you call them contributors with write permissions or repo maintainers, but it really seems like just one role. Either you have write perms or you don't. 
The root group maybe defines a more classical sense of 'maintainer', but again I'm not overly concerned with the labels.

vaporos: There are three distinct permissions here - review, merge, and approval. These three levels are defined here to reflect those three levels.  Review and merge is controlled via github permissions and approvers are probably in MAINTAINERS.md.  If commiters == maintainers in practice, great.  However, I suspect with more repos, we will have maintainers of some repos as committers (but not maintainers) for other repos.

text/0000-sawtooth-governance.md
vaporos: None of the Sawtooth components have a separate existence per-se, as it is one big project. We have different components but not different projects. Conceptually, something like Seth could be elevated to a proper Hyperledger project itself at some point.

Releases, publishing of artifacts, etc. need to be coordinated across the components. We could release components on their own release schedules (this makes a lot of sense for components which are not tightly coupled), but general strategy should roughly be the same (same tagging/branching approach, etc.).

For the SDKs, the sub-team will be represented by maintainers of all the SDKs (ideally) as it is important that we have similar features and design across the SDKs.

ineffectualproperty:
+1

aludvik:
After some further discussion, a few revisions have been made and the new proposed root team is as follows. This RFC will enter the Final Comment Period once all the proposed root team members below give explicit approval in this PR. Following approval from all members, this PR will be merged after 5 business days.

- [ ] Shawn Amundson
- [ ] Tom Barnes
- [ ] Andi Gunderson
- [ ] Adam Ludvik
- [ ] Dan Middleton
- [ ] James Mitchell
- [ ] Kelly Olson
- [ ] Peter Schwarz

aludvik:
This PR has entered the Final Comment Period and will be merged end-of-day Friday.

PR #7, opened 2018-04-23
Proposes Sawtooth Sabre transaction family

vaporos:
Proposes Sawtooth Sabre, a transaction family which implements on-chain
smart contracts executed in a WebAssembly virtual machine.

WebAssembly (Wasm) is a stack-based virtual machine newly implemented in
major browsers. It is well-suited for the purposes of smart contract
execution due to its sandboxed design, growing popularity, and tool
support.

Andi Gunderson wrote substantial portions of this RFC, including the
majority of the Reference-level explanation.

Signed-off-by: Shawn T. Amundson <amundson@bitwise.io>

### DIFF
--- a/text/0000-wasm-smart-contracts.md
+++ b/text/0000-wasm-smart-contracts.md
@@ -470,7 +470,6 @@ The inputs for CreateContractRegistryAction must include:
 The outputs for CreateContractRegistryAction must include:
 
 * the address for the contract registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### DeleteContractRegistryAction
 
@@ -498,7 +497,6 @@ The inputs for DeleteContractRegistryAction must include:
 The outputs for DeleteContractRegistryAction must include:
 
 * the address for the contract registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### UpdateContractRegistryOwnersAction
 
@@ -527,7 +525,6 @@ The inputs for UpdateContractRegistryOwnersAction must include:
 The outputs for UpdateContractRegistryOwnersAction must include:
 
 * the address for the contract registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### CreateNamespaceRegistryAction
 
@@ -556,7 +553,6 @@ The inputs for CreateNamespaceRegistryAction must include:
 The outputs for CreateNamespaceRegistryAction must include:
 
 * the address for the namespace registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### DeleteNamespaceRegistryAction
 
@@ -583,7 +579,6 @@ The inputs for DeleteNamespaceRegistryAction must include:
 The outputs for DeleteNamespaceRegistryAction must include:
 
 * the address for the namespace registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### UpdateNamespaceRegistryOwnersAction
 
@@ -612,7 +607,6 @@ The inputs for UpdateNamespaceRegistryOwnersAction must include:
 The outputs for UpdateNamespaceRegistryOwnersAction must include:
 
 * the address for the namespace registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### CreateNamespaceRegistryPermissionAction
 
@@ -647,7 +641,6 @@ The inputs for CreateNamespaceRegistryPermissionAction must include:
 The outputs for CreateNamespaceRegistryPermissionAction must include:
 
 * the address for the namespace registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ### DeleteNamespaceRegistryPermissionAction
 
@@ -674,7 +667,6 @@ The inputs for DeleteNamespaceRegistryPermissionAction must include:
 The outputs for DeleteNamespaceRegistryPermissionAction must include:
 
 * the address for the namespace registry
-* the settings address for ``sawtooth.swa.administrators``
 
 ## Transaction Header
 
--- a/text/0000-wasm-smart-contracts.md
+++ b/text/0000-wasm-smart-contracts.md
@@ -680,7 +680,7 @@ No dependencies.
 [subteam]: #subteam
 
 A Sawtooth Sabre sub-team shall be formed; this sub-team shall be responsible
-for the following related to Sabre: sheparding RFCs, accepting or rejecting
+for the following related to Sabre: shepharding RFCs, accepting or rejecting
 RFCs, policy decisions (including when RFCs are required), and making decisions
 on topics which do not require an RFC.
 
--- a/text/0000-wasm-smart-contracts.md
+++ b/text/0000-wasm-smart-contracts.md
@@ -63,8 +63,8 @@ SDK's transaction processor API. It further attempts to provide these
 capabilities in the most source-compatible way possible to make it easy to
 transition existing transaction processors to run as a Sabre smart contract.
 
-The goal of Sawtooth Sabre is to provide a Sawototh-native on-chain smart
-contract capbility to Sawtooth. The goal of the design and implementation
+The goal of Sawtooth Sabre is to provide a Sawtooth-native on-chain smart
+contract capability to Sawtooth. The goal of the design and implementation
 described in this RFC is to jumpstart Sabre with a minimal but usable
 implementation, not describe every eventual feature.
 
--- a/text/0000-wasm-smart-contracts.md
+++ b/text/0000-wasm-smart-contracts.md
@@ -89,23 +89,23 @@ The design of Sabre includes three fundamental data structures:
 
 Generally speaking, transaction processors can read and write arbitrary global
 state addresses. While restrictions can be placed upon the transaction
-processor via a couple different existing mechanisms, they are not granular
+processor via a couple of different existing mechanisms, they are not granular
 enough to be used in the context of Sabre, since all Sabre contracts would be
 subject to the same restrictions.
 
 Smart contracts and namespaces of global state are not necessarily a one-to-one
-relationship. For example, if we could create separate Sabre smart contracts
+relationship. For example, we could create separate Sabre smart contracts
 which provide multiplication and addition operations on the intkey namespace.
 Both smart contracts would operate on the same intkey namespace. Since
 this is the case, for a fully permissioned system we desire the ability to
 control access to the namespace independently from the permission to deploy and
 maintain the smart contracts.
 
-In Sabre, A namespace registry is used to control permissions related to
+In Sabre, a namespace registry is used to control permissions related to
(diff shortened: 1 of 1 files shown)


text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,770 @@
...
+[parity]:https://github.com/paritytech/parity/tree/27c32d3629e4b70b63ee37736147318035f48792/ethcore/wasm
+[wasmi]:https://github.com/paritytech/wasmi
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+The current design does not describe a method for limiting execution (in
+a manner similar to gas in Ethereum). It is therefore not yet suitable for
+deployment of code by potentially malicious actors. This will be addressed in
+a future RFC.
```
dcmiddle: Runaway code isn't restricted to just malicious actors. It can be the result of inadvertent programming errors. Even if it's not initially implemented, I think there either needs to be a premise how to do the execution accounting or a means of stopping a runaway contract.

vaporos: We don't handle this with transaction processors, so even without this feature designed and implemented, we are on-par with current state.

One possible way to implement this is to modify the bytecode and add the equivalent of a use_gas() call at the right points. Another would be to abort the interpreter purely based on the number of instructions it has executed. Currently, we have not taken a position on which path would be better.

vaporos:
Added additional commits which resolve typos, etc.

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,770 @@
...
+  }
+```
+
+If the contract registry for the provided contact name already exists, the
+transaction is invalid.
+
+The new contract registry is created for the name and provided owners. The
+owners should be a list of public keys of users that are allowed to add new
+contract versions, delete old versions, and delete the registry.
+
```
dcmiddle: Are there any permissioning possible on who can create a contract registry.

text/0000-wasm-smart-contracts.md
dcmiddle: I think the difference is that system admins need to take an active role in launching new transaction processors, and it's possible to gate adding TPs through the permissioning system. The concept of operations is that the consortium would review and ratify code before deploying it on the network.
Sabre would alleviate the logistical aspects of launching apps on the network but should offer some controls like the TP process to mitigate some risks. I added a question about permissioning earlier in the doc just now.
Permissioning might still be a bit less protection than status quo because a permissioned actor can still launch unreviewed code.
If the long term solution is instruction accounting, I would appreciate some dialog on the feasibility within wasm or some way to judge if we really have line of sight.

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,770 @@
...
+restricted by the Ethereum design and does not expose the entire feature set of
+Sawtooth transaction processors.
+
+Sawtooth Sabre fills this gap by providing a permissioned WebAssembly-based
+smart contract platform which provides full feature-parity with the Sawtooth
+SDK's transaction processor API. It further attempts to provide these
+capabilities in the most source-compatible way possible to make it easy to
+transition existing transaction processors to run as a Sabre smart contract.
+
+The goal of Sawtooth Sabre is to provide a Sawototh-native on-chain smart
```
agunde406: Sawototh-native  -> Sawtooth-native

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,770 @@
...
+contract capbility to Sawtooth. The goal of the design and implementation
```
agunde406: capbility -> capability

vaporos:
Added a commit to address typos found by @agunde406.

text/0000-wasm-smart-contracts.md
```diff
@@ -451,6 +451,11 @@ Create a contract registry with no version.
 If the contract registry for the provided contact name already exists, the
```
dcmiddle: typo?
s/contact/contract

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,785 @@
...
+
+If the contract registry for the contract name does not exist, the transaction
+is invalid.
+
+Both the new contract and the updated contract registry are set in state.
+
+The inputs for CreateContractAction must include:
+
+* the address for the new contract
+* the address for the contract registry
```
dcmiddle: Why is the address for the contract registry required? Is that implicit from the contract name? 
If the address is to be used to check validity, then the client shouldn't be allowed to submit it, the TP code should just compute it as ContractRegistry+hash(contract).

text/0000-wasm-smart-contracts.md
```diff
@@ -460,10 +465,12 @@ The new contract registry is set in state.
 The inputs for CreateContractRegistryAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
 
 The outputs for CreateContractRegistryAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
dcmiddle: does creating a contract registry modify the sawtooth.swa.administrators?

text/0000-wasm-smart-contracts.md
```diff
@@ -476,18 +483,22 @@ Deletes a contract registry if there are no versions.
...
 The inputs for DeleteContractRegistryAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
+
 
 The outputs for DeleteContractRegistryAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
dcmiddle: does deleting a contract modify the administrators?

text/0000-wasm-smart-contracts.md
```diff
@@ -501,18 +512,22 @@ Update the contract registry's owners list.
...
 The inputs for UpdateContractRegistryOwnersAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
+
 
 The outputs for UpdateContractRegistryOwnersAction must include:
 
 * the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
dcmiddle: update contract registry modifies administrators?

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,785 @@
...
+
+The inputs for CreateContractRegistryAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
+
+The outputs for CreateContractRegistryAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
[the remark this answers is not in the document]
agunde406: nope. This is an error. sawtooth.swa.administrators should be removed from the outputs.

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,785 @@
...
+The inputs for DeleteContractRegistryAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
+
+
+The outputs for DeleteContractRegistryAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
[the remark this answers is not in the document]
agunde406: Same as above

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,785 @@
...
+The inputs for UpdateContractRegistryOwnersAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
+
+
+The outputs for UpdateContractRegistryOwnersAction must include:
+
+* the address for the contract registry
+* the settings address for ``sawtooth.swa.administrators``
```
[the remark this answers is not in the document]
agunde406: same as above

vaporos:
Add a commit from @RyanLassigBanks to further describe limiting function execution.

Added commits from @agunde406 which address comments from @dcmiddle .

text/0000-wasm-smart-contracts.md
```diff
@@ -783,3 +783,8 @@ The current design does not describe a method for limiting execution (in
 a manner similar to gas in Ethereum). It is therefore not yet suitable for
 deployment of code by potentially malicious actors. This will be addressed in
 a future RFC.
+
+The interpreter used by Sabre, Wasmi, does not currently offer a way to halt
+the execution of a function. However, it would be relatively simple to add
+an option to the module constructor that would specify an upper limit for the
+number of intructions that can be executed.
```
aludvik: typo s/intructions/instructions/

vaporos:
Added a commit from @agunde406 but written by @aludvik to clarify the halting problem text.

text/0000-wasm-smart-contracts.md
```diff
@@ -0,0 +1,770 @@
...
+
+The use of named contracts is substantially different than other smart contract
+systems such as Ethereum, which instead use a hash of the contract as the
+unique identifier. The design was chosen because it provides a friendlier
+method of versioning smart contracts for permissioned networks. However, it
+would not be a suitable approach for a public non-permissioned network; thus,
+supporting such networks will require substantial design changes in the future.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
```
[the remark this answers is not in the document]
vaporos: Additional text related to this has been added to the Rationale and alternatives section

vaporos:
Propose this RFC be merged.

After approval by the root team, this RFC will enter its final comment period.

- [X] Shawn Amundson
- [ ] Tom Barnes
- [ ] Andi Gunderson
- [ ] Adam Ludvik
- [ ] Dan Middleton
- [ ] James Mitchell
- [ ] Kelly Olson
- [ ] Peter Schwarz

peterschwarz:
Before merging, you should be sure to add a commit that changes the file name to `0007-...` and add the appropriate links to the document.

ineffectualproperty:
Approved. As discussed in core-dev and per comments from @dcmiddle and @aludvik it would be good if future documentation could discuss how namespace-to-contract mappings can be made 1:1 and how namespace access and contract code can be made immutable e.g. by setting namespace or contract owner to null values. Only recommended as this is the status quo for smart contract systems like Ethereum.

vaporos:
Adding to what @ineffectualproperty suggests, in addition to the documentation on how to accomplish it with this design, we should remain open to changes either as a natural evolution of this design or alternate designs which we could adopt via the RFC process.  It may be good to document some use cases scenarios to use to judge such designs.

vaporos:
This RFC has entered final comment period, which will end May 18th, at which point this RFC will be merged.

PR #12, opened 2018-05-23
{Deprecated. Newer at PR#20} PoET 2.0 Consensus 

askmish:
Signed-off-by: Ashish Kumar Mishra <ashish.k.mishra@intel.com>

### DIFF
--- a/text/0001-poet2-consensus.md
+++ b/text/0001-poet2-consensus.md
@@ -1,0 +1,161 @@
+- Feature Name: PoET-2.0
+- Start Date: 2018-05-21
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+As blockchain technology matures, there is a pressing need for more efficient 
+consensus mechanisms that will scale to large networks without putting 
+potentially crippling demands on the infrastructure required to support it. 
+Proof of Elapsed Time (PoET) is an attempt at providing an innovative algorithm 
+that not only scales very efficiently, but also places very few demands on the 
+infrastructure.
+
+This document details a new mechanism for the PoET algorithm that overcomes 
+some of the challenges with the original algorithm.
+
+The intended audience of this document is architects, developers and anyone who 
+would like to get a better understanding of the details of the PoET 2.0 
+algorithm.
+
+This RFC describes a new PoET 2.0 Consensus for supporting the Intel® Software 
+Guard Extensions(SGX) machines without Platform Services Enclave(PSE).
+
+# Motivation
+[motivation]: #motivation
+
+PoET 1.0 offers a very unique and efficient way to achieve consensus based on 
+secure timers set inside an Intel® Software Guard Extensions (SGX) enclave. The 
+secure timer and the monotonic counters used in PoET 1.0 rely on 
+_SGX Platform Services_ to access these services from the system hardware. 
+_SGX Platform Services_, unfortunately, are not yet available on all SGX 
+enabled machines. While plans are in place for introducing these services 
+universally across the processor lineup, it will take some time to release the 
+services and for them to become ubiquitous. PoET 2.0 addresses these challenges 
+by making the PoET algorithm independent of the _SGX Platform Services_.
+
+# Definitions
+The following terms are used throughout the PoET spec and are defined here for 
+reference.
+
+*Enclave*
+
+A protected area in an application’s address space which provides 
+confidentiality and integrity even in the presence of privileged malware.
+
+The term can also be used to refer to a specific enclave that has been 
+initialized with a specific code and data.
+
+*EPID*
+
+An anonymous credential system. 
+
+*EPID Pseudonym*
+
+Pseudonym of an SGX platform used in linkable quotes. 
+It is part of the IAS attestation response according to IAS API 
+specifications. It is computed as a function of the service Basename 
+(validator network in our case) and the device’s EPID private key.
+
+*PPK, PSK*
+
+PoET ECC public and private key created by the PoET enclave.
+
+*IAS Report Key*
+
+IAS public key used to sign attestation reports as specified in the current 
+IAS API Guide.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In PoET 1.0, the PoET SGX enclave set a timer internally and would only release a WaitCertificate, once the timer elapsed. To be considered a candidate for consensus, a block (a.k.a Claim Block) would have to be accompanied by the corresponding WaitCertificate. This meant that peers receiving a Claim Block accompanied by the WaitCertificate, were assured that the block had not been forwarded prematurely, hence the phrase ‘Proof of Elapsed Time’.
+PoET 2.0, however, removes the reliance on internal timers and monotonic counters within SGX. This necessitates a different approach for enforcing wait times. Rather than set a timer inside SGX, PoET 2.0 relies on SGX to provide it with a duration to wait and sets the timer OUTSIDE SGX. A Claim Block is forwarded only after the timer has expired. The block is now accompanied by a WaitCertificate from SGX containing the duration that the block was expected to wait for.
+While this opens up the possibility of misuse by a malicious actor, the community of validators is engaged to enforce the wait on the block. Blocks arriving at peers earlier than suggested by their WaitCertificate are held back until the time is right to forward them. The larger the size of the network, the more resistant the protocol is to attack by a malicious actor or a cabal.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Here we describe the PoET 2.0 algorithm in greater detail.
+
+## Initialization
+When the PoET 2.0 module is initialized, it performs the following actions:
+1. Generate an ECDSA key pair. This pair is held in SGX memory and NEVER committed to disk for reuse subsequently
+2. Create a table in SGX memory to map the Block Number to WaitCertificates. Also store the block number of the chain head in SGX
+3. Generate a Quote containing the PPK and perform Quote verification with the IAS
+4. Check if there was a previous PPK registration. If there was, verify if ‘K’ blocks have been added to the chain since the key’s registration. If K blocks have not passed, wait until K blocks have been added. If the process restarts, signup at process startup and resume the K-test since this new signup
+5. Register the PPK and the IAS response with the ledger
+6. Wait for ‘C’ blocks to be added to the chain before participating in consensus(C-test)
+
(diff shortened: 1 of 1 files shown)


aludvik:
Quick general comment, I would be careful with your usage of the word "valid". I believe it is good to distinguish between a block's validity status and block's commit status. A block's validity should only depend on deterministic computation, things like signature verification, structure verification, batch validity relative to a state hash, etc. It should not be possible for a block's validity status to change over time. A block's committed status however can and should change over time based on the state of the network and other block's that have been considered.

To be more specific, I would not use "valid" in this sentence in the "Early Arriving blocks" section: `If the block is invalid i.e. it is an early-arriving block`. I would refer to block's that are dropped immediately because there is something intrinsically wrong with them (like a bad WaitCertificate) as invalid and refer to block's that do not have anything wrong with them but which we are not committing as something else, possibly "pending" or just "valid but not committed".

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+As blockchain technology matures, there is a pressing need for more efficient consensus mechanisms that will scale to large networks without putting potentially crippling demands on the infrastructure required to support it. Proof of Elapsed Time (PoET) is an attempt at providing an innovative algorithm that not only scales very efficiently, but also places very few demands on the infrastructure.
+This document details a new mechanism for the PoET algorithm that overcomes some of the challenges with the original algorithm.
+The intended audience of this document is architects, developers and anyone who would like to get a better understanding of the details of the PoET 2.0 algorithm.
+
+This RFC describes a new PoET 2.0 Consensus for supporting the SGX PSE-free machines.
```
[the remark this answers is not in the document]
chenette: And a minor note:  Best tech-writing practice is to use the full term "Intel® Software Guard Extensions (SGX)" here, because it's the first time the term is used.  But if that makes the sentence confusing or too long, it's probably okay to use "SGX" here, because the full term is in the next section.

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In PoET 1.0, the PoET SGX enclave set a timer internally and would only release a WaitCertificate, once the timer elapsed. To be considered a candidate for consensus, a block (a.k.a Claim Block) would have to be accompanied by the corresponding WaitCertificate. This meant that peers receiving a Claim Block accompanied by the WaitCertificate, were assured that the block had not been forwarded prematurely, hence the phrase ‘Proof of Elapsed Time’.
```
chenette: Nitpick: The term "claim block" is written inconsistently.  At first, it's "Claim Block". The rest of the paper uses "claim block", but there's also two occurrences of "ClaimBlock" without the space.

Technical writers prefer lower case (claim block), because it's easier to read. Plus, fewer terms would have to change. But "Claim Block" is acceptable as long as it's consistent.

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Wall Clock and Chain Clock
+The PoET 2.0 algorithm maintains two clocks, the _Wall Clock_ and the _Chain Clock_.
+The Wall Clock (WC) measures the time elapsed since the arrival of the ‘Sync Block’, i.e. the first block on the chain.
+The Chain Clock (CC) measures the cumulative wait duration of all the blocks in the chain (each fork will have its own CC).
+Upon receiving the Sync Block, WC and CC are initialized to 0. In practice, the WC may be calculated by recording the system time at the moment of the arrival of the Sync Block and subsequently subtracting this timestamp from the current time.
+To handle clock drifts, it may become necessary to synchronize with a standard network clock. Details are out of scope for this version of the document.
+
+## Block Creation
+Once the algorithm is in a state to publish blocks (i.e. the C test passes), it starts the process of creating claim blocks with the aim of participating in the Consensus mechanism.
```
chenette: I found "the C test" to be confusing.  Is it the same thing as Step 6 in the previous section? ("Wait for ‘C’ blocks to be added ...").  Could you clarify here?

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+The claim/candidate block is assembled as usual with no special processing required for PoET 2.0
+
+## Wait Certificate Creation
+Once a claim block is created, the PoET algorithm requests the PoET enclave to create the WaitCertificate.
+
+The contents of the WaitCertificate for PoET 2.0 are:
```
chenette: This information is hard to read in paragraph form.  I suggest using a bulleted list.  For example: 

     The contents of the WaitCertificate for PoET 2.0 are:

     * Duration: A random 256 bit number generated by SGX
     * WaitTime: The number of seconds to wait, determined by the LocalMean
     .
     .
     .

If you don't like bullets, you could simply add a blank line between each item so that each renders as its own paragraph.

The third choice is an RST term-definition list, but I always forget the syntax for that one.

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Block Publishing
+With the ClaimBlock ready and the WaitCertificate created, the validator will determine the duration for which to set the timer. The duration of the wait is determined by the ‘Duration’ field in the WaitCertificate.
+Ideally, any selected algorithm should be capable of converting the duration to a wait time distributed uniformly between 0 and LocalMean seconds.
+
+## Block Verification
+Upon receiving a claim block and a WaitCertificate, the receiving validator does the following:
+### Sanity checks
+1.	Ensure the Block and WaitCertificate signatures are valid
+2.	The Block’s ID matches the WaitCert’s blockID
```
chenette: Another nitpick: I suggest spelling out "WaitCertificate" (here and elsewhere) instead of abbreviating it in a few places. Consistency is good. :-)

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+15.	If the chain is switched, drop the block proposal & restart (Step 9).
+16.	Forward the new block over the gossip network.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+Poet 2.0 improves on PoET 1.0 adding support for PSE free devices. Current PoET 1.0 limits the availability of PoET consensus only to SGX PSE enabled devices.
```
chenette: As with my first comment above, I suggest using the term "(SGX) devices with/without Platform Services" (or something similar), instead of "PSE-free" or "PSE-enabled".

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+5.	Compute the new CC. The block is valid if CC + E < WC. Here E is a network latency coefficient added to simulate the average network delay seen by the validator.
+6.	Send the block to the validator for processing the transactions
+7.	If valid, commit the block, store the blockID and WaitCert in SGX
+8.	Forward the block to the gossip network
+9.	Drop the existing block proposal and restart from the new chain head.
+### Early Arriving blocks
+10.	If the block is invalid i.e. it is an early-arriving block, compare the duration with the duration of the claim block. If the block duration is smaller, drop the claim block. Wait for the WC to catch up with the CC. If the claim block duration is less than the incoming block duration, discard the incoming block.
+11.	Proceed to step 6 once the block becomes valid.
+### Fork Resolution
+12.	If the new block claims an earlier block as a parent, check for any cached blocks claiming the new block as a parent, retrieve them. Compute the new CC and check if the chain is valid (CC + E < WC)
```
rranjan3: Should not this be a recursive call until a terminal child is reached within the cache? There could be a fork comprising of multiple (>1 block) in the cache.

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,109 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Here we describe the PoET 2.0 algorithm in greater detail.
+
+## Initialization
+When the PoET 2.0 module is initialized, it performs the following actions:
+1. Generate an ECDSA key pair. This pair is held in SGX memory and NEVER committed to disk for reuse subsequently.
+2. Create a table in SGX memory to map the Block Number to WaitCertificates. Also store the block number of the chain head in SGX.
+3. Generate a Quote containing the ECDSA public key and perform Quote verification with the IAS.
```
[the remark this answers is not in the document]
askmish: The signup has to happen every time the process starts/system reboots based on the K-test, similar to what we have in poet 1.0.

text/0001-poet2-consensus.md
```diff
@@ -0,0 +1,161 @@
...
+6.	Send the block to the validator for processing the transactions
+7.	If valid, commit the block, store the blockID and WaitCertificate in SGX
+8.	Forward the block to the gossip network
+9.	Drop the existing block proposal and restart from the new chain head.
+### Early Arriving blocks
+10.	If the block is an early-arriving block, compare the duration with the duration of the Claim Block. If the block duration is smaller, drop the Claim Block. Wait for the WC to catch up with the CC. If the Claim Block duration is less than the incoming block duration, discard the incoming block.
+11.	Proceed to step 6 once the block becomes valid.
+### Fork Resolution
+12.	If the new block claims an earlier block as a parent, check for any cached blocks claiming the new block as a parent, retrieve them. Compute the new CC and check if the chain is valid (CC + E < WC)
+13.	If valid, compare existing chain length with the new chain length. If existing chain length is more, discard new chain. If the new chain is longer, switch to the new chain (commit the new blocks)
```
hartm: I have a question about this fork resolution process.  Can someone explain why we only need to compare chain lengths (rather than check the total number of "work" done on the chain--i.e., incorporate the population estimate)?

Here's my thought process:  suppose I controlled a small group of nodes.  We could could fork off everything to a side chain and then wait for the "difficulty" to go down.  In the steady state, due to the population adjustment, we would be adding blocks at the same rate as the main chain.  We could let this go on for a while--maybe we would get lucky and accumulate blocks at a slightly (sublinear) rate faster than the main chain.  Then, at some point, we could add in a bunch of new members and use these to add much more blocks than expected and try to catch up to the main chain.  It's possible there's some mathematical reasoning that says this is impossible, but it's not immediate (and needs to be written up if this is becomes the spec).

Does anyone have an explanation for why this doesn't work?  Am I misunderstanding something here?  Thanks!

askmish: @hartm  The newer spec for poet 2.0 is at https://github.com/hyperledger/sawtooth-rfcs/pull/20

PR #13, opened 2018-05-30
Expand available data types for Properties in Supply Chain

delventhalz:
This RFC proposes that the available data types used to describe Properties in Sawtooth Supply Chain should be expanded to include `BOOLEAN`, `ENUM`, and `STRUCT`. In addition,`FLOAT` should be removed and replaced by giving integers an exponent property that formalizes the practice of representing decimals as whole numbers of fractions (i.e. thousandths or millionths). This new more versatile integer property would be referred to as `NUMBER`.

As there was immediate need but no Supply Chain subteam, this RFC has already been implemented. Although this might seem to render the RFC pointless, it is not too late to make alterations to the existing implementation. The community should be given a chance to comment on this major update, and if changes are required, those should be implemented with further commits.

### DIFF
--- a/text/0000-supply-chain-expand-data-types.md
+++ b/text/0000-supply-chain-expand-data-types.md
@@ -1,0 +1,608 @@
+- Feature Name: supply_chain_expand_data_types
+- Start Date: 2018-03-08
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+The available "data types" used to describe Properties in Sawtooth Supply Chain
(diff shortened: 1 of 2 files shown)


delventhalz:
I propose this RFC enter the Final Comment Period. Approvals from the Supply Chain team?

- [X] Zac Delventhal @delventhalz
- [x] Shawn Amundson @vaporos 
- [x] James Mitchell @jsmitchell 
- [x] Kelly Olson @ineffectualproperty

PR #14, opened 2018-05-31
Replace existing Supply Chain clients with a new universal UI

delventhalz:
A proposal to develop a new Sawtooth Supply Chain client built to handle any use case, which will update and replace the existing FishNet and AssetTrack clients. This new client will have no hard coded references to any RecordTypes, and will dynamically populate pages like the Add Record form, using a new `/record-types` API route, and new RecordType properties stored in state.

Although specialized custom UIs are still likely to be used for production-level apps, having a single adaptable UI would be extremely useful both for experienced developers looking to quickly stand-up up a demo with new data, or new developers looking for one canonical UI to reference.


delventhalz:
I propose this RFC enter the Final Comment Period. Waiting on approvals from the remainder of the Supply Chain team.

- [x] Zac Delventhal @delventhalz
- [x] Shawn Amundson @vaporos 
- [] James Mitchell @jsmitchell 
- [] Kelly Olson @ineffectualproperty

delventhalz:
This proposal to remove FishNet and AssetTrack and replace them with a single universal client has entered the final comment period. If there are no substantive changes requested by July 26, it will be merged.

PR #15, opened 2018-05-31
Store Supply Chain Property schema on chain as PropertyTypes

delventhalz:
A proposal to store some PropertySchemas permanently on chain as a `PropertyType`. These type schema can then be referenced by name when creating a new RecordType. This will copy the schema into the RecordType, as well as save the name with the actual Properties for clients to reference.

This will allow multiple RecordTypes to share are reuse the concept of particular properties. For example, "temperature" could be defined once, and every use on every type of record would be consistent. Furthermore, since clients can read these types, they could be more sophisticated about how they parse properties (for example, displaying every temperature with a thermometer instead of just a number).

### DIFF
--- a/text/0000-supply-chain-property-references.md
+++ b/text/0000-supply-chain-property-references.md
@@ -1,0 +1,322 @@
+- Feature Name: supply_chain_property_references
+- Start Date: 2018-03-14
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+Store PropertySchemas on-chain. These schema can then be referenced by name
+when creating a new RecordType. This will copy the schema into the RecordType,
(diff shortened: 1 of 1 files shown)


delventhalz:
I propose this RFC enter the Final Comment Period. Waiting on approvals from the remainder of the Supply Chain team.

- [x] Zac Delventhal @delventhalz
- [x] Shawn Amundson @vaporos 
- [ ] James Mitchell @jsmitchell 
- [ ] Kelly Olson @ineffectualproperty

delventhalz:
This proposal to add PropertyTypes to Sawtooth Supply Chain has entered the final comment period. If there are no substantive requests for changes by August 2, 2018 it will be merged.

PR #23, opened 2018-08-03
Allow transaction processor to get transaction header serialized bytes in order to validate signature

yoni-wolf:


vaporos:
Please wrap the text at less than 80 columns and fill out the remaining sections omitting any example text. To pass the DCO check, sign your git commits and re-push. The specific implementation change will need to be described in detail and an explanation of why the change is backward compatible.

I'll add per-line comments after those changes.

A possible implementation approach would be to modify TpRegisterRequest to indicate that the transaction processor desires header bytes.

```
message TpRegisterRequest {
    message TpProcessRequestHeaderStyle {
        STYLE_UNSET,
        EXPANDED,
        RAW
    }

    string family = 1;
    string version = 2;
    repeated string namespaces = 4;
    uint32 max_occupancy = 5;
    TpProcessRequestHeaderStyle process_request_header_style;
}
 ```

With a new field within TpRegisterRequest:

```
message TpProcessRequest {
    TransactionHeader header = 1;  // The transaction header
    bytes payload = 2;  // The transaction payload
    string signature = 3;  // The transaction header_signature
    string context_id = 4; // The context_id for state requests.
    bytes header_raw = 5;
}
```

For EXPANDED, 'header' field would be filled in, for RAW, 'header_raw' would be filled in.

Other topics should be discussed as well, such as why it is not sufficient to do this within the payload itself.

File: text/0000-raw-txn-header.md [hunk 1]
```diff
@@ -0,0 +1,96 @@
...
+in protobuf that de-serializing and re-serializing back will produce same bytes.
+
+A more reliable soultion is for the validator to forward the serialized header
+bytes of the transaction object, now calling the new get_serialized_header()
+will provide the bytes without the need to re-serialize. For this recommended
+solution, there is also an option to add flag to TP registration in order to let
+TP decide if it requires this API.
+
+A possible implementation approach would be to modify TpRegisterRequest to
+indicate that the transaction processor desires header bytes.
```
vaporos: Because of the nature of this change (changing an API we have committed to backwards compatibility), this should be restated in more firm terms omitting language like "possible", since this is a concrete proposal.  Alternatives should be added to the "rationale and alternatives" section.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+[drawbacks]: #drawbacks
+
+Most transaction processors do not need the raw header bytes.
+Enabling this feature may mean carrying extra data through the sawtooth-core
+internal pipeline.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+An alternative would be to include a signature within the payload itself. That
```
vaporos: This is technically impossible, since the signature is generated from the payload's hash. You could store a hash of a portion of the payload with in the payload.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+field.
+
+# Prior art [prior-art]: #prior-art
+
+This problem appears unique to the conjunction of the sawtooth architecture and
+the use of a trusted execution environment for the transaction handling logic.
+
+# Unresolved questions [unresolved]: #unresolved-questions
+
+Internal sawtooth-core management of the header is not prescribed in this RFC.
```
vaporos: I think you mean 'validator' instead of 'sawtooth-core'.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+[alternatives]: #alternatives
+
+An alternative would be to include a signature within the payload itself. That
+has the downside of bloating the transactions with an additional redundant
+field.
+
+# Prior art [prior-art]: #prior-art
+
+This problem appears unique to the conjunction of the sawtooth architecture and
+the use of a trusted execution environment for the transaction handling logic.
```
vaporos: Actually, 0.8 had an API which sent the raw bytes, so that's the best prior art to point to. This API changed during 1.0 stabilization to send the transaction payload as it is now.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+[summary]: #summary
+
+Requesting to add API to the sdk - txn.get_serialized_header(), that will return
+the transaction header bytes.
+
+# Motivation 
+[motivation]: #motivation
+
+For security reasons transaction processor would like to verify the incoming
+transaction header, and signature.  For example, in the private ledger TP (C++)
```
vaporos: Is the "private ledger TP" available to link to? If not, it should not be mentioned here as readers will not have the appropriate context.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+that runs inside SGX enclave, we can't trust the signature verification done by
+sawtooth since it happens outside of SGX.  In order to verify the transaction
+signature, TP needs to get the serialized transaction header and not just all
+the fields.
```
vaporos: This does not explain why it is important to be able to verify the signature within the TEE.

It seems like this could have been solved entirely within the payload content.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+    bytes header_raw = 5;
+}
+For EXPANDED, 'header' field would be filled in, for RAW, 'header_raw' would
+be filled in.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Most transaction processors do not need the raw header bytes.
+Enabling this feature may mean carrying extra data through the sawtooth-core
```
vaporos: Suggest instead "Adding this feature will increase the stable API surface for the benefit of a single transaction processor implementation. (Though in theory it could be useful to other transaction processors in the future, we do not know of any plans.) The stable API is important because we have made a commitment of backward support, so this feature, if added, will need to be supported into the future.

This will also increase the complexity of the validator slightly."

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
+- Feature Name: send_serialize_txn
+- Start Date: 2018-08-03
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+Requesting to add API to the sdk - txn.get_serialized_header(), that will return
```
vaporos: We may want this to be get_raw_header().

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
+- Feature Name: send_serialize_txn
```
vaporos: Change the feature name here to raw_txn_header to match the filename.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,96 @@
...
+
+For security reasons transaction processor would like to verify the incoming
+transaction header, and signature.  For example, in the private ledger TP (C++)
+that runs inside SGX enclave, we can't trust the signature verification done by
+sawtooth since it happens outside of SGX.  In order to verify the transaction
+signature, TP needs to get the serialized transaction header and not just all
+the fields.
+
+
+# Guide-level explanation
```
vaporos: Either in the guide or reference explanation, we should have an example of the API change (maybe by showing the lines required within apply() to get to the raw header. Ideally, this example would be given in Rust but Python would be okay as well.

text/0000-raw-txn-header.md
yoni-wolf: payload is verified with its hash that is in the signed txn header, TP needs a way to detect that attacker didn't replace the txn payload from different txn, only way to know that is if we check signature of header that includes the payload hash

dcmiddle:
Thanks for the updates @yoni-wolf. I think you may have a patch ready to go for this. If so could you link to a branch with that change here?

yoni-wolf:
attaching, please note this is an implementation i made for our project that only applies for sawtooth c++ sdk, it is not the implementation as requested in this RFC
[serialized_header_cxx.txt](https://github.com/hyperledger/sawtooth-rfcs/files/2486118/serialized_header_cxx.txt)

vaporos:
I am generally supportive of this RFC. We need more scrutiny by the core team with respect to the API changes proposed and backward compatibility concerns.

An observation - we only have a single use case for this, which could probably be solved by separately hashing the contents within the payload:

```
message Payload {
    string contents_sha512 = 1;
    bytes contents = 2;
}
```

Though this particular approach wouldn't cover the header information, that could be included easy enough (without the payload hash, and causing duplication). It would also require application-level support, which is probably its biggest flaw.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+An alternative would be to sign the payload itself. That has the downside of 
+bloating the transactions with an additional redundant field from the header.
+
+# Prior art [prior-art]: #prior-art
+
+sawtooth 0.8 had had an API which sent the raw bytes, This API changed during
+1.0 stabilization to send the transaction payload as it is now
+Missing this API appears unique to the conjunction of the sawtooth
+architecture and the use of a trusted execution environment for the 
+transaction handling logic.
```
aludvik: Please clean up the above paragraph. There are several typos and incorrect uses of punctuation, and the last sentence is a little confusing.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
+- Feature Name: raw_txn_header
+- Start Date: 2018-08-03
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+Requesting to add API to the sdk - txn.get_raw_header()(), that will return
```
aludvik: typo, `()` appears twice

aludvik:
I am good with this.

text/0000-raw-txn-header.md
aludvik: The transaction processor API in Hyperledger Sawtooth v0.8 sent the raw header bytes, however this API was changed during the 1.0 stabilization period to its current state. The requirement that raw header bytes be included in the transaction processor API seems to be unique to processors that use a trusted execution environment.

yoni-wolf:
thanks for the review @agunde406 and @aludvik, committed changes based on your comments

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+No change is recommended to the existing methods, so implementation of this RFC
+should not cause any breakage for pre-existing transaction processors.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The validator to forward the serialized header bytes of the transaction
+object, now calling the new get_serialized_header() will provide the
+bytes without the need to re-serialize. 
+there is also an option to add flag to TP registration in order to let TP
```
jsmitchell: Sentence fragment

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC adds a new method to directly access the transaction header bytes.
+Transaction Processor developers will use this API when they choose to re-verify
+the signature inside of a transaction processor.
+
+No change is recommended to the existing methods, so implementation of this RFC
```
jsmitchell: s/so implementation/so the implementation/

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+should not cause any breakage for pre-existing transaction processors.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The validator to forward the serialized header bytes of the transaction
```
jsmitchell: I suggest replacing this sentence with something like: "Calling get_serialized_header() will return the serialized bytes of the transaction object."

text/0000-raw-txn-header.md
jsmitchell: s/add flag/add a flag/

jsmitchell: /to TP registration/to the TP registration/

jsmitchell: s/to let TP/to let the TP/

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+            EXPANDED,
+            RAW
+        }
+
+        string family = 1;
+        string version = 2;
+        repeated string namespaces = 4;
+        uint32 max_occupancy = 5;
+        TpProcessRequestHeaderStyle process_request_header_style = 6;
+    }
```
jsmitchell: Insert line feed after }

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+With a new field within TpRegisterRequest:
+
+    message TpProcessRequest {
+        TransactionHeader header = 1;  // The transaction header
+        bytes payload = 2;  // The transaction payload
+        string signature = 3;  // The transaction header_signature
+        string context_id = 4; // The context_id for state requests.
+        bytes header_raw = 5;
+    }
```
jsmitchell: Insert line feed after }

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+plans.) The stable API is important because we have made a commitment of 
+backward support, so this feature, if added, will need to be supported into
+the future.
+
+This will also increase the complexity of the validator slightly.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+An alternative would be that txn object that arrives to TP in 'apply' method
```
jsmitchell: s/txn/transaction/

jsmitchell: s/that arrives to TP in 'apply' method/which is provided to the TP's 'apply' method/

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+the future.
+
+This will also increase the complexity of the validator slightly.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+An alternative would be that txn object that arrives to TP in 'apply' method 
+will have an api to re-serialize the transaction header. 
+This solution could be problematic since it requires trust in protobuf that
```
jsmitchell: s/requires trust/requires placing trust/

jsmitchell: s/produce same bytes/produce the same bytes/

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+backward support, so this feature, if added, will need to be supported into
+the future.
+
+This will also increase the complexity of the validator slightly.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+An alternative would be that txn object that arrives to TP in 'apply' method 
+will have an api to re-serialize the transaction header.
```
jsmitchell: s/api/API/

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
+- Feature Name: raw_txn_header
+- Start Date: 2018-08-03
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+Requesting to add API to the sdk - txn.get_raw_header(), that will return
```
jsmitchell: Replace this sentence with something like "This RFC proposes an additional API call, txn.get_raw_header(), which will return the transaction's header bytes."

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,103 @@
...
+the signature inside of a transaction processor.
+
+No change is recommended to the existing methods, so implementation of this RFC
+should not cause any breakage for pre-existing transaction processors.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The validator to forward the serialized header bytes of the transaction
+object, now calling the new get_serialized_header() will provide the
```
jsmitchell: Is it txn.get_serialized_header() or txn.get_raw_header() ?

yoni-wolf:
Thank you @jsmitchell for the review, committed changes based on your comments

agunde406:
@yoni-wolf your new commit is missing a sign-off

dcmiddle:
> Propose this RFC be merged. After approval by the core subteam, this RFC will enter its final comment period.
> 
> * [ ]  Adam Ludvik @aludvik
> * [ ]  Andi Gunderson @agunde406
> * [x]  Dan Middleton @dcmiddle
> * [x]  James Mitchell @jsmitchell
> * [ ]  Peter Schwarz @peterschwarz
> * [ ]  Shawn Amundson @vaporos

Re-initiating request for vote following most recent updates.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,104 @@
...
+
+    message TpProcessRequest {
+        TransactionHeader header = 1;  // The transaction header
+        bytes payload = 2;  // The transaction payload
+        string signature = 3;  // The transaction header_signature
+        string context_id = 4; // The context_id for state requests.
+        bytes header_raw = 5;
+    }
+
+For EXPANDED, 'header' field would be filled in, for RAW, 'header_raw' would
```
peterschwarz: What is the behavior when this is not set?  For backwards compatibility, I would suggest that the default should be expanded.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,104 @@
...
+single transaction processor implementation. (Though in theory it could be 
+useful to other transaction processors in the future, we do not know of any
+plans.) The stable API is important because we have made a commitment of 
+backward support, so this feature, if added, will need to be supported into
+the future.
+
+This will also increase the complexity of the validator slightly.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
```
peterschwarz: Would an additional alternative be to implement a `Signer` that executes in the TEE?

yoni-wolf: You mean that sawtooth will validate the transaction in TEE?
Attacker could still hack ZMQ and send a different transaction to TP, a secure TP must validate any data that comes from outside TEE.
Only option i can think of is that validator and TP both run in TEE and communicate directly enclave to enclave (which also requires some kind of communication authentication)

text/0000-raw-txn-header.md
yoni-wolf: Agree, update RFC to reflect this

agunde406:
You are missing sign off on one of your commits. Also, the merge commits should not be included. Instead, you should rebase your branch on master.

dcmiddle:
@aludvik just need your check / approval.

text/0000-raw-txn-header.md
```diff
@@ -0,0 +1,105 @@
...
+object without the need to re-serialize.
+There is also an option to add a flag to the TP registration in order to let the 
+TP decide if it requires this API.
+
+A possible implementation approach would be to modify TpRegisterRequest to
+indicate that the transaction processor desires header bytes.
+
+    message TpRegisterRequest {
+        message TpProcessRequestHeaderStyle {
+            STYLE_UNSET,
```
arsulegai: [Question]: Default is EXPANDED, do we still need STYLE_UNSET? If so, in which case?
Apologies if I am missing something here and for late commenting on it.

PR #29, opened 2018-10-23
PBFT Regular view changes

aludvik:
Describes two extensions to Sawtooth PBFT for handling liveness and fairness problems.

### DIFF
--- a/text/0000-pbft-regular-view-change.md
+++ b/text/0000-pbft-regular-view-change.md
@@ -38,18 +38,19 @@ more resilient to bad actors.
 
 This RFC also proposes a mitigating solution to the "silent leader" problem.
 Sawtooth PBFT checks for leader liveness by starting a timer when a leader
-proposes a new block. If the timer expires before the new block is committed,
-the leader is suspected of being faulty and a view change is initiated.
-However, if a leader in the NotStarted state never proposes a block, no timer
-is started. This means that a faulty leader can "remain silent" and stall the
+proposes a new block and endorses it with a pre-prepare message. If the timer
+expires before the new block is committed, the leader is suspected of being
+faulty and a view change is initiated. However, if a leader never proposes a
+block or doesn't endorse a block with a pre-prepare message, no timer is
+started. This means that a faulty leader can "remain silent" and stall the
 network without the other nodes being able to determine whether the silence is
 because no new batches are arriving, or the leader is intentionally ignoring
 them.
 
-                NotStarted  <----------- Finished
+                PrePreparing  <----------- Finished
     (start timer) |                           ^
                   v                           | (stop timer)
-                PrePreparing -> ... -> Committing
+                Preparing -------------> Committing
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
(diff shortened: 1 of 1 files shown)


text/0000-pbft-regular-view-change.md
```diff
@@ -0,0 +1,185 @@
...
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC describes an extension to the PBFT implementation to mitigate the
+effect of compromised leader nodes. The RFC takes into account leaders unfairly
+ordering batches within blocks (the "unfair ordering" problem) and leaders
+intentionally not producing new blocks to stall the network (the "silent
+leader" problem).
```
dcmiddle: The second problem is addressed in the original Castro paper. Is this proposal intended to implement that solution or provide an different solution (see section 4.4 of Castro)?

text/0000-pbft-regular-view-change.md
```diff
@@ -0,0 +1,185 @@
...
+
+In the second case, nodes propose a standard view change when the network is
+idle for too long (the primary does not produce a block and a pre-prepare
+message). This mitigates the "silent leader" problem by limiting the amount of
+time a faulty leader can stall the network.
+
+The setting `sawtooth.consensus.pbft.forced_view_change_period` determines how
+often, measured in blocks, a view change should be forced. After transitioning
+from the Finished to PrePreparing phase, a node will check whether a view change
+should be forced and, if so, immediately change views. Specifically, the node
```
dcmiddle: My read of the Castro paper is that the backup proposing a view change will be the leader (primary) of the new view. Which node would be elected in this case? Would it be deterministic, and if so does that predictability present a new and meaningful risk?

text/0000-pbft-regular-view-change.md
ltseeley: Section 4.4 of the paper ("View Changes") does not actually solve this problem. "View
changes are triggered by timeouts that prevent backups from waiting indefinitely for requests to execute." This timeout only starts once a request is received (in our case a block) and makes sure that it will execute at some point. It doesn't handle the case where no request (block) ever gets received.

text/0000-pbft-regular-view-change.md
ltseeley: That is incorrect; the new primary is deterministically calculated based on the ordering of the nodes (in the `sawtooth.consensus.pbft.peers` setting) and the view number. If you have _n_ nodes that are numbered _0 .. n - 1_, node _p_ will be the primary as determined by the formula: _p = v mod n_, where _v_ is the view number.

ltseeley: So basically, when you force a view change (increment the view by 1), it will just be the next node in the on-chain list that becomes the primary.

agunde406:
Propose this RFC be merged. After approval by the core subteam, this RFC will enter its final comment period.

  - [ ] Shawn Amundson
  - [x] Andi Gunderson
  - [ ] Adam Ludvik
  - [ ] Dan Middleton
  - [ ] James Mitchell
  - [ ] Peter Schwarz

text/0000-pbft-regular-view-change.md
dcmiddle: Ok. So long as we are addressing the client's request (i.e. transactions by a transactor) I think that is the spirit of requests in the paper. The way I read the paper, the lead validator is aggregating requests into blocks and then transmitting that aggregation of requests (the block) in the subsequent stages (preprepare, prepare, commit). The reply stage should probably then disaggregate the block to message the original clients.

text/0000-pbft-regular-view-change.md
dcmiddle: Yes, I was incorrect. The sequencing you describe is consistent with the paper.

agunde406:
This RFC has entered the final comment period. If there are no substantive requests for changes by March 29, 2019 it will be merged.

PR #30, opened 2018-10-23
PBFT Consensus seal

aludvik:


agunde406:
Propose this RFC be merged. After approval by the core subteam, this RFC will enter its final comment period.

  - [ ] Shawn Amundson
  - [x] Andi Gunderson
  - [ ] Adam Ludvik
  - [ ] Dan Middleton
  - [ ] James Mitchell
  - [ ] Peter Schwarz

dcmiddle:
I previously approved this, but I feel like there was some conversation about using aggregated signatures that either didn't happen or wasn't captured. 
In short, I still agree that as written this would be a useful feature, but I would encourage any developers to consider aggregated signatures in lieu of separate signature fields appended for the n replica votes.

ltseeley:
> I previously approved this, but I feel like there was some conversation about using aggregated signatures that either didn't happen or wasn't captured.
> In short, I still agree that as written this would be a useful feature, but I would encourage any developers to consider aggregated signatures in lieu of separate signature fields appended for the n replica votes.

Can you elaborate on what you mean by "aggregated signatures?" I don't recall this coming up in previous discussions, but maybe that's because it wasn't captured.

agunde406:
This RFC has entered the final comment period. If there are no substantive requests for changes by April 1st, 2019 it will be merged.

dcmiddle:
> > I previously approved this, but I feel like there was some conversation about using aggregated signatures that either didn't happen or wasn't captured.
> > In short, I still agree that as written this would be a useful feature, but I would encourage any developers to consider aggregated signatures in lieu of separate signature fields appended for the n replica votes.
> 
> Can you elaborate on what you mean by "aggregated signatures?" I don't recall this coming up in previous discussions, but maybe that's because it wasn't captured.

In a nutshell the signatures are added together so you end up carrying a single signature in the final block rather than n signatures. Similarly you verify by adding up the public keys and the aggregated signature will only pass if it is the combination of all the expected signatures.

PR #32, opened 2018-11-13
Add new peers without restarting a node

eugene-babichenko:
Proposes an extension of validator client messaging to add peers to a running node without restarting it.

Signed-off-by: Yevhenii Babichenko <eugene@remme.io>

### DIFF
--- a/text/0000-add-peers-in-runtime.md
+++ b/text/0000-add-peers-in-runtime.md
@@ -1,0 +1,173 @@
+- Feature Name: `add_peers_in_runtime`
+- Start Date: 2018-11-12
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+This RFC proposes to implement the possibility to add new peers in runtime
+(through the component validator endpoint) when a node is working in the
+`static` peering mode. Along with that it also adds the corresponding extensions
+to the off-chain permissioning model.
+
+# Motivation
+[motivation]: #motivation
+
+When an administrator adds a new node to an existing Sawtooth network he/she has
+to restart a node with new peering settings. This makes any automation
+significantly harder to write than if we had the possibility to add peers in the
+runtime and also decreases the uptime.
+
+To resolve this problem our team proposes to add a method to add new peers to a
+running node.
+
+An example use case is using Sawtooth along with a service discovery system like
+[Consul](https://www.consul.io):
+
+- A Consul node is set up along with the Sawtooth node;
+- A middleware continuously fetches the changes of the peer list from the
+  Consul node;
+- The middleware adds peers to the Sawtooth node via the validator component
+  endpoint.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When an administrator adds new peers to the network you very likely want to
+add them without restarting the validator with a newer `--peers` parameter
+value. To add a new peer you can send the `ClientAddPeersRequest` to the
+`component` endpoint of your Sawtooth validator.
+
+The Protocol Buffers definition for this message is the following:
+
+```protobuf
+message ClientAddPeersRequest {
+    repeated string peers = 1;
+}
+```
+
+Simple Python example using `sawtooth_sdk.messaging.stream`:
+
+```python
+new_peers = ['tcp://192.168.0.100:8008', 'tcp://192.168.0.200:8008']
+add_peers_request = ClientAddPeersRequest(new_peers=new_peers)
+future = stream.send(Message.CLIENT_ADD_PEERS_REQUEST,
+                     add_peers_request.SerializeToString())
+response_serialized = future.result().content
+response = ClientAddPeersResponse()
+response.ParseFromString(response_serialized)
+```
+
+The response Protocol Buffers definition is the following:
+
+```protobuf
+message ClientAddPeersResponse {
+    enum Status {
+        STATUS_UNSET = 0;
+        OK = 1;
+        INTERNAL_ERROR = 2;
+        // One or more of peer URIs were malformed. List of malformed URIs is in
+        // the `invalid_uris` field of this response.
+        INVALID_PEER_URI = 3;
+        MAXIMUM_PEERS_CONNECTIVITY_REACHED = 4;
+    }
+    Status status = 1;
+    repeated string invalid_uris = 2;
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Protocol Buffers definitions
+[protobuf]: #protobuf
+
+I propose to add them to `protos/client_peers.proto`:
+
+```protobuf
+message ClientAddPeersRequest {
+    repeated string peers = 1;
+}
+
+message ClientAddPeersResponse {
+    enum Status {
+        STATUS_UNSET = 0;
+        OK = 1;
+        INTERNAL_ERROR = 2;
+        // One or more of peer URIs were malformed. List of malformed URIs is in
+        // the `invalid_uris` field of this response.
+        INVALID_PEER_URI = 3;
+        MAXIMUM_PEERS_CONNECTIVITY_REACHED = 4;
+    }
+    Status status = 1;
+    repeated string invalid_uris = 2;
+}
+```
+
+The rationale behind the `invalid_uris` is to be more precise about what is
+wrong and to ease the debugging process for developers.
+
+## How are the requests processed by the validator
+[request-processing]: #request-processing
+
+The requests are received on the `component` endpoint. When the validator
+receives a new request for adding peers it:
+
+- Validates the format of peer URIs which has to be
+  `tcp://IP_ADDRESS:PORT_NUMBER`
+- If the validation was successful then the validator updates its peer list and
+  immediately returns the `OK` response. The new peers are connected _after_
+  that.
+
+Edge cases:
+
+- The peer address format was wrong in one or more of the provided peers. If
+  that happens then the request fails without adding any new peers to the peers
+  list and returns the `INVALID_PEER_URI` status along with the list of faulty
+  peer URIs.
+- If the `--maximum-peer-connectivity` parameter was provided to the validator
+  then the validator checks if it has reached the maximum peer connectivity and
+  fails with an error if so. The validator also fails if it cannot add _all_ of
+  the peers provided in a request without breaking the provided
+  `maximum-peer-connectivity`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- The proposed solution does not provide any information on the status of new
+  peers. It just returns immediately if a request does not break the conditions
+  specified in the previous chapter.
+- It does not specify any connection retry policies leaving it to the existing
+  peering implementation.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+Alternatives were not considered and judging from multiple examples this is the
+state-of-the-art solution.
+
+# Prior art
+[prior-art]: #prior-art
+
+- [`addnode` in Bitcoin JSON RPC](https://bitcoincore.org/en/doc/0.16.0/rpc/network/addnode/)
+- [`admin_addPeer` in Ethereum management API](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#admin_addpeer)
+
+Those two allow adding new peers in their platforms. Interesting points:
+
+- Ethereum can return the connection status;
+- Bitcoin allows specifying the connection retry policy.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- During the pre-RFC discussion on the #sawtooth-core-dev channel, there was no
+  final solution on "should it be included to the REST API or no?"
+- In the same discussion, there was a proposition to resolve security issues by
+  using this feature along with the permissioning module. If we do not add this
+  feature to the REST API and hence to the administrative utilities then I do
+  not see any point in permissioning because the described feature remains an
+  internal interface of the application. Even if we do then we can restrict the
+  access to that feature by using a proxy as suggested in the documentation.
+- Should the system notify its user about the statuses of new connections or
+  leave the status check to the end user?
+- Should our team include the integration example of this solution for Consul?


text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,188 @@
...
+# Motivation
+[motivation]: #motivation
+
+When an administrator adds a new node to an existing Sawtooth network he/she has
+to restart a node with new peering settings. This makes any automation
+significantly harder to write than if we had the possibility to add peers in the
+runtime and also decreases the uptime.
+
+To resolve this problem our team proposes to add a method to add new peers to a
+running node.
```
dcmiddle: Have you also considered removing nodes? The security implications are a little more risky but about the same. The benefit of including a remove function would let us address the limits of maximum-peer-connectivity.

dcmiddle:
I think this would be a useful addition to the administrative APIs. Main thing to work out are security cosiderations.

text/0000-add-peers-in-runtime.md
agunde406: I agree that it would be good to include removal functionality for discussion .

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+## How are the requests processed by the validator
+[request-processing]: #request-processing
+
+The requests are received on the `component` endpoint. When the validator
+receives a new request for adding peers it:
+
+- Validates the format of peer URIs which has to be
+  `tcp://IP_ADDRESS:PORT_NUMBER`
```
aludvik: In addition to IP addresses, this should also accept DNS names, which ZMQ will resolve: http://api.zeromq.org/2-1:zmq-tcp

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,188 @@
...
+- If the validation was successful then the validator updates its peer list and
+  immediately returns the `OK` response. The new peers are connected _after_
+  that.
```
aludvik: If the response is sent back before connecting, there is no way to notify clients of authorization errors. Is not sending an error due to authorization an intentional decision? Or should the error enum be extended and a response sent only after peering completes?

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,188 @@
...
+- During the pre-RFC discussion on the #sawtooth-core-dev channel, there was no
+  final solution on "should it be included to the REST API or no?"
+- In the same discussion, there was a proposition to resolve security issues by
+  using this feature along with the permissioning module. If we do not add this
+  feature to the REST API and hence to the administrative utilities then I do
+  not see any point in permissioning because the described feature remains an
+  internal interface of the application. Even if we do then we can restrict the
+  access to that feature by using a proxy as suggested in the documentation.
+- Should the system notify its user about the statuses of new connections or
+  leave the status check to the end user?
```
aludvik: My opinion would be to either have the response only be sent after the new connection is created, or expose another request type that allows the user to poll on the status of the new connection so they can detect connection failures such as authorization errors.

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,188 @@
...
+- Start Date: 2018-11-12
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+This RFC proposes to implement the possibility to add new peers in runtime
+(through the component validator endpoint) when a node is working in the
+`static` peering mode. Along with that it also adds the corresponding extensions
+to the off-chain permissioning model.
```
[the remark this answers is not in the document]
eugene-babichenko: Okay, after this explanation I have a better understanding of the permissioning requirement. Will describe a model that was discussed on the chat.

text/0000-add-peers-in-runtime.md
eugene-babichenko: I think that this is a good idea.

text/0000-add-peers-in-runtime.md
eugene-babichenko: > Is not sending an error due to authorization an intentional decision

I missed it intentionally to have a discussion on how we should handle such errors.

In my opinion, it would be better to extend the enum in `ClientAddPeersResponse` but then I would like to discuss the implementation. The simplest approach I see here is to make the [`Gossip`](https://github.com/hyperledger/sawtooth-core/blob/ffad8e38d9c26b278b374bc5edaf776e5d0edcb2/validator/sawtooth_validator/gossip/gossip.py#L77) class send notifications about peer status changes. This will allow us to know if the connection attempt was ok or not ok, but will not give any specific information. To be more specific we will have to add handlers for corresponding peer messages that will notify us, for example, about authorization violations, rejected connections and so on. I think that the simplest approach should be acceptable now (at least in the system our company build).

eugene-babichenko:
Just uploaded an update. Here is what is changed:

- Extended the range of error codes for adding peers.
- Added the description for peers removal.
- Now requests for adding peers are submitted one by one instead of doing that in batches. The rationale behind that is fairly simple. Now this RFC support error codes that are returned after the connection is finished or failed. Because of that we need control over each separate connection and they may fail with different errors.
- Added the description of the permissioning model. This is very basic and I doubt if it should be generalized somehow (added that to the "Unresloved questions" section).

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,210 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+When an administrator adds a new node to an existing Sawtooth network he/she has
+to restart a node with new peering settings. This makes any automation
+significantly harder to write than if we had the possibility to add peers in the
+runtime and also decreases the uptime.
+
+To resolve this problem our team proposes to add a method to add new peers to a
```
agunde406: Please add remove as well to this sentence

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,210 @@
...
+    enum Status {
+        STATUS_UNSET = 0;
+        OK = 1;
+        INTERNAL_ERROR = 2;
+        ADMIN_AUTHORIZATION_ERROR = 3;
+        // One or more of peer URIs were malformed. List of malformed URIs is in
+        // the `invalid_uris` field of this response.
+        INVALID_PEER_URI = 4;
+        MAXIMUM_PEERS_CONNECTIVITY_REACHED = 5;
+        AUTHORIZATION_VIOLATION = 6;
```
agunde406: Can you explain why you have AUTHORIZATION_VIOLATION here along with ADMIN_AUTHORIZATION_ERROR, but only ADMIN_AUTHORIZATION_ERROR in ClientRemovePeerResponse?

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,210 @@
...
+We should also add new message types to `protos/validator.proto`:
+
+```protobuf
+message Message {
+
+    enum MessageType {
+        // ...
+        CLIENT_PEERS_ADD_REQUEST = 131;
+        CLIENT_PEERS_ADD_RESPONSE = 132;
+        CLIENT_PEERS_REMOVE_REQUEST = 131;
```
agunde406: This should be 133

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,210 @@
...
+        CLIENT_PEERS_REMOVE_RESPONSE = 132;
```
agunde406: This should be 134

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,210 @@
...
+    }
+    // ...
+}
+```
+
+## How are the requests processed by the validator
+[request-processing]: #request-processing
+
+The requests are received on the `component` endpoint. When the validator
+receives a new request for adding peers it:
```
agunde406: Please also add what the validator would do to remove peers

text/0000-add-peers-in-runtime.md
eugene-babichenko: `AUTHORIZATION_VIOLATION` is the error thrown during the process of establishing a new connection from validator to another and is definitely irrelevant in `ClientRemovePeerResponse`.

text/0000-add-peers-in-runtime.md
eugene-babichenko: Done.

text/0000-add-peers-in-runtime.md
eugene-babichenko: Done.

text/0000-add-peers-in-runtime.md
eugene-babichenko: Done.

text/0000-add-peers-in-runtime.md
eugene-babichenko: Added that, and there is one question: what to do if we go below the minimum peer connectivity limit here? I have two options in my mind:

- Just allow to do that, but it my bring the node functionality down;
- Another option is to throw an error here in case we reached the minimum limit.

Wondering which of those two will be a better solution.

text/0000-add-peers-in-runtime.md
agunde406: Ah i see

text/0000-add-peers-in-runtime.md
agunde406: Isnt this proposal only for static peering? minimum peer connectivity is used for dynamic peering.

eugene-babichenko: My bad, sorry

eugene-babichenko:
I am not sure if we should define a separate namespace here &mdash; this falls beyond the scope of this proposal. I followed the simple logic here: if something does not belong to TPs or consensus engine, this should be in the Client* namespace.

Do you mean by "to be bound to a different port" that those messages should be on a different port than `component`?

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
+- Feature Name: `add_peers_in_runtime`
+- Start Date: 2018-11-12
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary This RFC proposes to implement the possibility to add new
+peers and remove the existing connections in runtime (through the component
```
jsmitchell: Recommend: "This RFC proposes functionality to add and remove static peer connections while a validator node is running."

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+validator endpoint) when a node is working in the `static` peering mode. Along
+with that it also adds the corresponding extensions to the off-chain
```
jsmitchell: Recommend: "This RFC also adds the corresponding extensions to the off-chain permssioning model."

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+permissioning model.
+
+# Motivation
+[motivation]: #motivation
+
+When an administrator adds a new node to an existing Sawtooth network he/she has
```
jsmitchell: add comma after 'network' and just pick either he or she for the gender

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+to restart a node with new peering settings. This makes any automation
+significantly harder to write than if we had the possibility to add peers in the
```
jsmitchell: Recommend: "Restarting the process adds substantial complexity to infrastructure automation, and incurs system downtime."

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+runtime and also decreases the uptime.
+
+To resolve this problem our team proposes to add a method to add new peers to a
```
jsmitchell: Add a comma after problem

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+- A Consul node is set up along with the Sawtooth node;
+- A middleware continuously fetches the changes of the peer list from the Consul
+  node;
+- The middleware adds peers to the Sawtooth node via the validator component
+  endpoint.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When an administrator adds new peers to the network you very likely want to add
```
jsmitchell: Recommend: "When an administrator adds new peers to the network, she likely wants to connect to them without needing to restart existing validators with an updated `--peers` parameter value."

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+them without restarting the validator with a newer `--peers` parameter value. To
+add a new peer you can send the `ClientAddPeersRequest` to the `component`
```
jsmitchell: Recommend: "To add a new peer, the administrator can send..." and replace 'your' with 'their' later in the sentence.

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+response.ParseFromString(response_serialized)
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Protocol Buffers definitions
+[protobuf]: #protobuf
+
+I propose to add them to `protos/client_peers.proto`:
```
jsmitchell: Remove I/we language

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+        PEER_NOT_FOUND = 4;
+    }
+    Status status = 1;
+}
+```
+
+The rationale behind the `invalid_uris` is to be more precise about what is
+wrong and to ease the debugging process for developers.
+
+We should also add new message types to `protos/validator.proto`:
```
jsmitchell: Remove I/we language

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+[request-processing]: #request-processing
+
+The requests are received on the `component` endpoint.
+
+### Adding peers
+
+When the validator receives a new request for adding peers it:
+
+- Validates the format of peer URI which has to be `tcp://ADDRESS:PORT_NUMBER`;
+- If the validation was successful then the validator tries to connect to a
```
jsmitchell: Add comma after successful

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  provided peer. If the connection was successful it returns the `OK` status.
```
jsmitchell: Add comma after successful

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  Otherwise the corresponding error status is returned.
+
+Edge cases:
+
+- The peer address format was wrong in one or more of the provided peers. If
+  that happens then the request fails without adding any new peers to the peers
```
jsmitchell: Add comma after happens

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+The requests are received on the `component` endpoint.
+
+### Adding peers
+
+When the validator receives a new request for adding peers it:
+
+- Validates the format of peer URI which has to be `tcp://ADDRESS:PORT_NUMBER`;
+- If the validation was successful then the validator tries to connect to a
+  provided peer. If the connection was successful it returns the `OK` status.
+  Otherwise the corresponding error status is returned.
```
jsmitchell: Add comma after Otherwise

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+
+Edge cases:
+
+- The peer address format was wrong in one or more of the provided peers. If
+  that happens then the request fails without adding any new peers to the peers
+  list and returns the `INVALID_PEER_URI` status along with the list of faulty
+  peer URIs.
+- If the `--maximum-peer-connectivity` parameter was provided to the validator
```
jsmitchell: Add comma after "provided to the validator"

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  then the validator checks if it has reached the maximum peer connectivity and
```
jsmitchell: Add comma after connectivity

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  the peers provided in a request without breaking the provided
+  `maximum-peer-connectivity`.
+
+### Removing peers
+
+When the validator receives a new request for removing peers it does the
+following:
+
+- Validates the format of peer URI which has to be `tcp://ADDRESS:PORT_NUMBER`;
+- If a peer is connected the validator removes it. Otherwise the
```
jsmitchell: Add comma after connected

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  `PEER_NOT_FOUND` error is thrown.
+
+## Permissioning
+
+The proposition is to add the `admin` role to off-chain permissioning that will
+restrict access to requests that can be malicious. The workflow for the
+validation is the following:
+
+- If the `admin` role is not specified then the permissioning module will use
```
jsmitchell: Add comma after specified

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  the `default` policy.
+- If the `default` policy is not specified then the validation of the
```
jsmitchell: Add comma after specified

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+  permissions is not performed and fields `admin_public_key` and `signature` can
+  be omitted.
+- If the `default` or the `admin` policy is specified, then the permission
+  verifier checks:
+  - If the `admin_public_key` is allowed.
+  - If the `signature` is correct.
+  - If one of the above conditions is not satisfied then the
```
jsmitchell: Add comma after satisfied

text/0000-add-peers-in-runtime.md
```diff
@@ -0,0 +1,222 @@
...
+
+- If the `admin` role is not specified then the permissioning module will use
+  the `default` policy.
+- If the `default` policy is not specified then the validation of the
+  permissions is not performed and fields `admin_public_key` and `signature` can
+  be omitted.
+- If the `default` or the `admin` policy is specified, then the permission
+  verifier checks:
+  - If the `admin_public_key` is allowed.
+  - If the `signature` is correct.
```
vaporos: Can you expand on this? In particular, it is not clear what is being signed.

vaporos:
Should the authorization for this feature occur during the connection conversation (see authorization.proto)? I think it may be appropriate to add ADMIN to "RoleType" and create an Admin* message namespace in addition to Client*, etc.

text/0000-add-peers-in-runtime.md
eugene-babichenko: This is the signature of the `peer_uri` string (https://github.com/hyperledger/sawtooth-rfcs/pull/32/files#diff-60b9c6f8741b42854f2726db45c3f899R94).

eugene-babichenko:
> Should the authorization for this feature occur during the connection conversation (see authorization.proto)?

It may be difficult to implement the authorization in the component endpoint. Should we extend the component endpoint, use the network endpoint (it already has the authorization implemented) or create a separate admin endpoint?

> I think it may be appropriate to add ADMIN to "RoleType" and create an Admin* message namespace in addition to Client*, etc.

Agree on that.

text/0000-add-peers-in-runtime.md
dcmiddle: The operation should also be covered by the signature. If only the URI is signed then it could be replayed in a ClientRemovePeerRequest. You could look at the batch format for an example of using a signature that covers the entirety of the message. https://github.com/hyperledger/sawtooth-core/blob/master/protos/batch.proto
There may be a simpler approach for this kind of message.

eugene-babichenko: I am still waiting for a reply to [this message](https://github.com/hyperledger/sawtooth-rfcs/pull/32#issuecomment-484020989). I am not really sure whether we should use just message signatures or develop a more complicated system for future use (I believe that such system should appear as a separate and more general RFC).

dcmiddle:
> > Should the authorization for this feature occur during the connection conversation (see authorization.proto)?
> 
> It may be difficult to implement the authorization in the component endpoint. Should we extend the component endpoint, use the network endpoint (it already has the authorization implemented) or create a separate admin endpoint?

I think an admin endpoint would be the most secure. This would facilitate firewall rules to restrict traffic. The downside is it adds one more element to a node setup. We used to get a lot of support questions stemming from configuring the existing endpoints. That said, I don't think this will add much complexity. A loopback default wouldn't need to be modified. A simple setup would then have the administrator logging into the validator host to issue these commands and the respective private key could reside on that validator host adding no other key management burden.

eugene-babichenko:
Is that OK if I reopen this pull request from another repository?

agunde406:
Please dont, we dont want to loose this conversation. @eugene-babichenko

eugene-babichenko:
@agunde406 The problem is I no longer have access to this fork but still want to finish this RFC.

agunde406:
@eugene-babichenko Ah okay. In that case, yes it is okay to resubmit the RFC from a new fork. Please do not change commit history and link this PR from the new one.

eugene-babichenko:
@agunde406 Here is the new pull request https://github.com/hyperledger/sawtooth-rfcs/pull/44

PR #39, opened 2019-02-21
delete by prefix rfc

yoni-wolf:
Signed-off-by: Yoni <yoni.wolf@intel.com>

### DIFF
--- a/text/0000-delete-prefix.md
+++ b/text/0000-delete-prefix.md
@@ -1,0 +1,135 @@
+- Feature Name: delete_by_prefix
+- Start Date: 2019-2-20
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+This RFC proposes an additional API to transaction context,
+context.deletePrefix(address_prefix) which will get an address prefix as input
+parameter and delete all the addresses in the prefix subtree.
+
+# Motivation
+[motivation]: #motivation
+
+The motivation came when looking for an efficient way to delete chunk of 
+addresses, for example all addresses that are allocated for a certain user.
+If each member has some unique address prefix, deleting all the addresses 
+belong to a member can become more efficient with delete by prefix.
+
+Since addresses are stored in a radix tree, deleting many addresses by one 
+prefix is much faster than deleting the same amount of addresses one by one.
+Furthermore, delete by prefix doesn't require transaction processor to know 
+which addresses exist. This will remove the need to store huge list of 
+existing addresses in a special address that will need to be modified by every
+transaction that writes a new address.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A context API to delete address already exist for transaction processors.
+The existing delete API accepts only full 35 bytes addresses.
+This RFC exposes to transaction processor a new capability of deleting by 
+address prefix. When delete by prefix is called, all addresses under the 
+prefix subtree will be deleted.
+
+Main usage for this feature is the need to support some 'clean transaction'
+that will remove all addresses with some communality, for example all 
+addresses that belong to some member.
+The proposed solution is to have address mapping logic that allocates some
+unique pre-defined address prefix per member. The transaction processor logic
+will know to calculate the address suffix based on transaction payload data
+and execute any transaction.
+In order to execute the clean transaction, the transaction processor will
+call context.deletePrefix with the member prefix.
+
+The existing solution is to keep track of all used addresses, this will
+require read-modify-write of a special address for every new address that is
+used. This solution could work but will slow down executing of every regular 
+transactions while the clean transaction is rare.
+
+No change is recommended to the existing methods, so the implementation of this
+RFC should not cause any breakage for pre-existing transaction processors.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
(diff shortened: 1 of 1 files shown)


text/0000-delete-prefix.md
```diff
@@ -0,0 +1,135 @@
+- Feature Name: delete_by_prefix
+- Start Date: 2019-2-20
+- RFC PR:
+- Sawtooth Issue:
+
+# Summary
+[summary]: #summary
+
+This RFC proposes an additional API to transaction context,
+context.deletePrefix(address_prefix) which will get an address prefix as input
```
arsulegai: Can we extend existing API and allow partial address?

yoni-wolf: We could.
I thought in order to prevent TP developers from deleting prefix by mistake, the right approach will be to add a new API for this instead of adding capabilities to existing delete API.
Do you think it will be better to allow delete with prefix by extending existing API?

arsulegai: Yes, this can however be controlled by outputs field in TransactionHeader.

text/0000-delete-prefix.md
```diff
@@ -0,0 +1,135 @@
...
+Addition to state_context.proto:
+	
+	// A request from the handler/tp to delete state prefix entries at an 
+	// collection of addresses prefix subtree
+	message TpStateDeletePrefixRequest {
+		string context_id = 1;
+		repeated string addresses_prefix = 2;
+	}
+
+	// A response form the contextmanager/validator with the addresses that
```
agunde406: form -> from

agunde406: contextmanager -> context_manager

yoni-wolf: thank you, fixed

text/0000-delete-prefix.md
```diff
@@ -0,0 +1,135 @@
...
+	// A response form the contextmanager/validator with the addresses that 
+	// were deleted
+	message TpStateDeleteResponse {
+		enum Status {
+			STATUS_UNSET = 0;
+			OK = 1;
+			AUTHORIZATION_ERROR = 2;
+		}
+
+		repeated string addresses = 1;
```
peterschwarz: I'm not really sure if it it's necessary (or desirable) to return the list of addresses. It has the same performance problem of the previously suggested RFC for getting all the set addresses out of state.  These addresses would need to be read under that prefix, and then the prefix can be deleted. 

Deleting the prefix is cheap. Knowning what is under the prefix is not.

text/0000-delete-prefix.md
```diff
@@ -0,0 +1,135 @@
...
+This has a performance downside that each time transaction processor needs to 
+write a new address, it will need to modify the special address containing 
+list of all addresses. This address could be huge and it needs to be changed
+for almost every transaction, while the need to delete all the addresses in 
+the list is very rare.
+
+# Prior art
+[prior-art]: #prior-art
+
+Access address by prefix is a key capability of radix tree.
```
peterschwarz: This is worded badly. In fact, I don't know that I would go so far as to even include prior art, since it is an extension of current Sawtooth - the Merkle-Radix trie - which that functionality made use of prior art.

text/0000-delete-prefix.md
```diff
@@ -0,0 +1,135 @@
...
+	// A request from the handler/tp to delete state prefix entries at an 
+	// collection of addresses prefix subtree
+	message TpStateDeletePrefixRequest {
+		string context_id = 1;
+		repeated string addresses_prefix = 2;
+	}
+
+	// A response from the context_manager/validator with the addresses
+	// that were deleted
+	message TpStateDeleteResponse {
```
agunde406: This message does not include the addresses that were deleted as stated above by the comment.

yoni-wolf: thank you, fixed

text/0000-delete-prefix.md
yoni-wolf: agree, removed the list

text/0000-delete-prefix.md
vaporos: What would extending the existing API look like?

vaporos:
In several places it looks like two paragraphs are touching and it's not clear if they are meant to be separate paragraphs or whether it's one paragraph with sloppy line wrapping.

In several places Sawtooth should be capitalized.

text/0000-delete-prefix.md
dcmiddle: @yoni-wolf please correct me, but I think that would look like the existing state delete call would take a list of address prefixes rather than a list of addresses. The validator would then need to distinguish each element of the list as between a full 70 character address as a single address or something less than 70 as a prefix.
Combining this: https://github.com/hyperledger/sawtooth-core/blob/master/protos/state_context.proto#L66-L70
With this: https://github.com/hyperledger/sawtooth-core/blob/master/protos/state_context.proto#L66-L70
@vaporos if that's not what you were asking please elaborate.

dcmiddle: Clarification was that this option should be included in the alternatives section.

yoni-wolf: Added to the alternative section the option to use existing API instead of adding a new API

PR #45, opened 2019-06-19
Sawtooth RFC for Docker Build and Execution in a Proxy Environment

danintel:
Signed-off-by: danintel <daniel.anderson@intel.com>

This PR submits the Docker proxy RFC after final reviews and modifications by everyone interested with the Google Doc:

https://docs.google.com/document/d/1Xux_qbx8edu5_bWYZsnfr7p5JoC3ZpPCaoDhhykChno/edit

### DIFF
--- a/text/0045-docker-proxy.md
+++ b/text/0045-docker-proxy.md
@@ -1,0 +1,148 @@
+- Feature Name: docker_proxy
+- Start Date: 2019-06-19
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Sawtooth has containerized build and test processes for core and all components. Running Docker
+containers behind a network proxy requires special handling so that the containers can find
+external routes to download packages and build dependencies, for example. There are a number of
+ways to configure these environment for proxy support. It is desirable that we adopt a single
+consistent approach across the large body of Docker and Docker Compose files in the Sawtooth
+project.
+
+This RFC defines the canonical approach for supporting proxies for Sawtooth containers using
+Docker 18.x with Ubuntu 16.04 and 18.04 and serves as guidelines to resolve these issues.
+
+# Motivation
+[motivation]: #motivation
+
+Sawtooth repositories make extensive use of Docker containers to build and test the software as
+well as provide turn key examples. Those containers typically need to fetch external packages or
+make other outbound network calls. The containers need special configuration to be executed behind
+network proxies. There are multiple ways to provide that proxy configuration and it is desirable
+that the Sawtooth repositories standardize on a single approach. The challenge is further
+complicated by different abilities and constraints of different versions of Docker and Ubuntu.
+
+Starting with Ubuntu 18 (Bionic), `gpg` , invoked through `apt-key adv` , now ignores `http_proxy`
+and `https_proxy` environment variables, and ignores the command line options
+`apt-key adv --keyserver-options http-proxy=<proxy server url>`
+and `https-proxy=<proxy server url>` . The only method remaining is with cumbersome
+apt command-specific configuration file settings in `/etc/apt/apt.conf.d/proxy` .
+
+
+# Goals
+[goals]: #goals
+
+* Have Docker and Docker Compose files and shell scripts that work in explicit proxy,
+  transparent proxy, and non-proxy environments
+* Have a uniform method to access proxy servers across Hyperledger Sawtooth projects
+* Minimize Docker and script clutter with unnecessary tests and settings
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In order to use the proposed containers, developers and users in proxied environments must set
+these environment variables:
+
+* If you use unsecure HTTP web access, set `http_proxy` . The setting is site-specific for your
+  corporate proxy server. For example,
+  `export http_proxy=http://proxy.foocorp.com:8080`
+* Similarly, if you use secure HTTPS/SSL/TLS web access, set `https_proxy` . The port number or
+  server or both are usually different from `http_proxy`
+* Similarly set `ftp_proxy` if you use FTP (this is not common)
+* Set `no_proxy` to a comma-separated list of internal corporate subnets. For example,
+  `no_proxy=.foocorp.com,10.0.0.0/8,localhost,127.0.0.0/8`
+* Do *not* set `HTTP_PROXY` , `HTTPS_PROXY` , and `NO_PROXY` . Most or all software use the
+  lower case `*_proxy` environment variables and only sometimes use UPPER CASE `*_PROXY` variables
+* Proxy support should be briefly mentioned in the README.md file or similar file, along with the
+  environment variable(s) needed, such as `HTTP_PROXY`
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+All Docker files must be updated to conform as follows:
+* Commands `apt, apt-get, cargo, deb, go, npm, pip3, wget` and others require no changes
+* `curl` requires no changes, but it is not usually installed. It needs to be installed for
+  containers that use it. For example,
+  `RUN apt-get update && apt-get install -yq curl `
+
+* `apt-key` : Instead of using `apt-key adv` to download an apt signing key, use `curl` and
+  `apt-key add -` . For example, instead of the following in a `Dockerfile` (which no longer
+  works):
+
+```
+RUN \
+ apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
+ --recv-keys 8AA7AF1F1091A5FD
+```
+
+Use the following, which works in proxy (with http_proxy) and non-proxy environments:
+
+```
+RUN \
+ apt-get update && apt-get install -yq curl && curl -sSL \
+ 'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD’\
+ | apt-key add -
+```
+
+All Docker Compose files must, pass `http_proxy` , `https_proxy` , and `no_proxy` through to each
+container that needs network access, and exclude the Docker network. For example:
+
+```
+my-sawtooth-client:
+  container_name: my-sawtooth-client
+  build:
+    context: .
+    dockerfile: ./my-sawtooth-client/Dockerfile
+    args:
+      - http_proxy
+      - https_proxy
+      - no_proxy
+  environment:
+    - 'http_proxy=${http_proxy}'
+    - 'https_proxy=${https_proxy}'
+    - 'no_proxy=rest-api,validator,${no_proxy}'
+```
+
+Instructions in `build.md` files must be updated accordingly. 
+
+# Conclusion
+[conclusion]: #conclusion
+
+These recommendations allow Docker and Docker compose use with Sawtooth software in a standard,
+concise manner that work in both a corporate proxy and non-proxy (or transparent proxy)
+environments.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The current alternative used is hand-modifying Docker scripts each time a proxy is required.
+This is time consuming, expensive, and unnecessary. 
+
+An earlier alternative of using `apt-key adv` and passing proxy variables on the gpg command line
+(discussed above) worked in Ubuntu Xenial, but not in Bionic. This alternative has the
+disadvantage of making Docker scripts messy with a lot of proxy configuration setup.
+
+A third alternative, setting configuration for newer versions of gpg (in Bionic, but not in
+Xenial) is more difficult as it does not work in Xenial and requires adding GPG-specific
+configuration files containing proxy configuration (instead of just passing an environment
+variable. This will also require messy proxy configuration setup in Docker.
+
+
+# Prior art
+[prior-art]: #prior-art
+
+Not applicable.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None.
--- a/text/0045-docker-proxy.md
+++ b/text/0045-docker-proxy.md
@@ -58,45 +58,9 @@ In order to use the proposed containers, developers and users in proxied
 environments must set these proxy variables in their
 `$HOME/.docker/config.json` configuration file. The `config.json` file
 resides on the Docker host system, not in the Docker container filesystem.
-The relevant settings are:
 
-* `httpProxy` sets the proxy and port for unsecure web (HTTP) requests.
-  The setting is site-specific for your corporate proxy server. For example,
-  `"httpProxy": "http://proxy.foocorp.com:8080",`
-* `httpsProxy` sets the proxy and port for HTTPS (TLS) requests. For example,
-  `"httpsProxy": "http://proxy.foocorp.com:4443",`
-* `ftpProxy` sets the proxy for FTP file transfer (this is not common)
-* `noProxy` sets a comma-separated list of networks and hosts that are in
-  the internal network. For example,
-  `"noProxy": "192.168.0.0/16,172.16.0.0/12,10.0.0.0,localhost,127.0.0.0/8,::1,foocorp.com",`
-
-  Make sure you include Docker Compose internal network(s), which by default
-   are within `172.16.0.0/12` (use `netstat -rn` to list your networks)
-* Do not set `http_proxy` , `https_proxy` , and `no_proxy` or their upper
-  case equivalents. This method has been replaced with the `config.json`
-  settings.
-* Proxy support should be briefly mentioned in the README.md file or similar
-  file, along with the proxy variable(s) needed, such as `httpProxy`
-
-The following is a generic example of `$HOME/.docker/config.json` . The
-actual contents are site specific:
-
-```
-{
- "proxies":
- {
-   "default":
-   {
-     "httpProxy": "http://proxy.foocorp.com:8080",
-     "httpsProxy": "http://proxy.foocorp.com:4443",
-     "noProxy": "192.168.0.0/16,172.16.0.0/12,10.0.0.0,localhost,127.0.0.0/8,::1,foocorp.com"
-   }
- }
-}
-````
-
-For more detail, see the section "Configure the Docker client" for Docker 17.07
-and higher in "Configure Docker to use a proxy server" 
+The file format is documented at in the section "Configure the Docker client"
+for Docker 17.07 and higher in "Configure Docker to use a proxy server" at
 https://docs.docker.com/network/proxy/#configure-the-docker-client
 
 # Reference-level explanation
@@ -108,7 +72,7 @@ All Docker files must be updated to conform as follows:
 * `curl` requires no changes, but it is not usually installed. It needs to be
   installed for containers that use it. For example,
   `RUN apt-get update && apt-get install -yq curl `
-
+* Package `gnupg2` is required by `apt-key` and should be installed.
 * `apt-key` : Instead of using `apt-key adv` to download an apt signing key,
   use `curl` and `apt-key add -` . For example, instead of the following in a
   `Dockerfile` (which no longer works):
@@ -125,7 +89,7 @@ two keyservers for robustness:
 
 ```
 RUN \
- apt-get update && apt-get install -yq curl && curl --retry 3 -sSL \
+ apt-get update && apt-get install -yq curl gnupg2 && curl --retry 3 -sSL \
     'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD' \
     | apt-key add - 
 ```
--- a/text/0045-docker-proxy.md
+++ b/text/0045-docker-proxy.md
@@ -1,77 +1,110 @@
 - Feature Name: docker_proxy
-- Start Date: 2019-06-19
-- RFC PR: (leave this empty)
-- Sawtooth Issue: (leave this empty)
+- Start Date: 2019-07-01
+- RFC PR: https://github.com/hyperledger/sawtooth-rfcs/pull/45
+- Sawtooth Issue: N/A
 
 # Summary
 [summary]: #summary
 
-Sawtooth has containerized build and test processes for core and all components. Running Docker
-containers behind a network proxy requires special handling so that the containers can find
-external routes to download packages and build dependencies, for example. There are a number of
-ways to configure these environment for proxy support. It is desirable that we adopt a single
-consistent approach across the large body of Docker and Docker Compose files in the Sawtooth
-project.
+Sawtooth has containerized build and test processes for core and all
+components. Running Docker containers behind a network proxy requires
+special handling so that the containers can find external routes to download
+packages and build dependencies, for example. There are a number of
+ways to configure these environment for proxy support. It is desirable that we
+adopt a single consistent approach across the large body of Docker and
+Docker Compose files in the Sawtooth project.
 
-This RFC defines the canonical approach for supporting proxies for Sawtooth containers using
-Docker 18.x with Ubuntu 16.04 and 18.04 and serves as guidelines to resolve these issues.
+This RFC defines the canonical approach for supporting proxies for
+Sawtooth containers using Docker 18.x and Docker Compose 1.24.x with
+Ubuntu 16.04 and 18.04 and serves as guidelines to resolve these issues.
 
 # Motivation
 [motivation]: #motivation
 
-Sawtooth repositories make extensive use of Docker containers to build and test the software as
-well as provide turn key examples. Those containers typically need to fetch external packages or
-make other outbound network calls. The containers need special configuration to be executed behind
-network proxies. There are multiple ways to provide that proxy configuration and it is desirable
-that the Sawtooth repositories standardize on a single approach. The challenge is further
-complicated by different abilities and constraints of different versions of Docker and Ubuntu.
+Sawtooth repositories make extensive use of Docker containers to build and
+test the software as well as provide turn key examples. Those containers
+typically need to fetch external packages or make other outbound network
+calls. The containers need special configuration to be executed behind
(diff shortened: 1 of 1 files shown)


text/0045-docker-proxy.md
```diff
@@ -0,0 +1,115 @@
+- Feature Name: docker_proxy
+- Start Date: 2019-06-19
+- RFC PR: (leave this empty)
+- Sawtooth Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Sawtooth has containerized build and test processes for core and all components. Running Docker containers behind a network proxy requires special handling so that the containers can find external routes to download packages and build dependencies, for example. There are a number of ways to configure these environment for proxy support. It is desirable that we adopt a single consistent approach across the large body of Docker and Docker Compose files in the Sawtooth project.
```
agunde406: Please wrap the lines to 80 characters

danintel: Lines wrapped.

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,148 @@
...
+Use the following, which works in proxy (with http_proxy) and non-proxy environments:
+
+```
+RUN \
+ apt-get update && apt-get install -yq curl && curl -sSL \
+ 'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD’\
+ | apt-key add -
+```
+
+All Docker Compose files must, pass `http_proxy` , `https_proxy` , and `no_proxy` through to each
```
arsulegai: This depends on the docker version. Maybe we can mention that Sawtooth assumes user is on latest version and in version 18.03 onwards this explicit passing of env is not required.

danintel: I didn't know that. I will verify the *_proxy variables are passed to Docker and update the document.

danintel: I tried a simple test that removed the `*_proxy` variables from a Docker file and the application now fails.  I used my CookieJar application with docker 18.09.6 and docker-compose 1.21.0.  Here is the diff:

```
$ git diff
diff --git a/docker-compose.yaml b/docker-compose.yaml
index 934b1ee..8ac2031 100644
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -39,10 +39,6 @@ services:
         - http_proxy
         - https_proxy
         - no_proxy
-    environment:
-      - 'http_proxy=${http_proxy}'
-      - 'https_proxy=${https_proxy}'
-      - 'no_proxy=rest-api,validator,${no_proxy}'
     volumes:
       - '.:/project/cookiejar/'
     depends_on:
```

The appliation worked before the change. After rebuilding the docker container with the above change, this is the result:

```
$ sudo docker exec -it cookiejar-cl
ient bash
[sudo] password for dano: 
root@e18112d57b62:/project/cookiejar/pyclient# cookiejar.py bake 100
URL to send to REST API is http://rest-api:8008/batches
[18:47:26 INFO    connectionpool] Starting new HTTP connection (1): proxy-jf.intel.com
[18:47:26 DEBUG   connectionpool] "POST http://rest-api:8008/batches HTTP/1.1" 404 654
Traceback (most recent call last):
  File "/project/cookiejar/pyclient/cookiejar_client.py", line 137, in _send_to_rest_api
    result.status_code, result.reason))
Exception: Error 404: Not Found
```

docker and docker-compose version:
```
 $ docker version
Client:
 Version:           18.09.6
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        481bc77
 Built:             Sat May  4 02:35:57 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.6
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       481bc77
  Built:            Sat May  4 01:59:36 2019
  OS/Arch:          linux/amd64
  Experimental:     false
daniela1-test1 sawtooth-cookiejar (docker) $ docker compose version
docker: 'compose' is not a docker command.
See 'docker --help'
daniela1-test1 sawtooth-cookiejar (docker) $ docker-compose version
docker-compose version 1.21.0, build 5920eb0
docker-py version: 3.2.1
CPython version: 3.6.5
OpenSSL version: OpenSSL 1.0.1t  3 May 2016
```

arsulegai: There are two things here:
1. If you pruned all docker images and rebuilt them by removing `args` under the `build` then `config.json` worked as expected. Proxy env variables were passed by docker during build time.
2. These env are also passed during container runtime, hence you saw see the error while executing here. By adding `rest-api` to the `no_proxy` list in `config.json` this issue would be resolved.

We're on same page, how do we handle `no_proxy` is the next question.

danintel: Sorry, I still do not understand what's being proposed here. This `config.json` file. Does it go in the host? Where? `/etc/docker/config.json` ? Or does it go in each of the Docker containers? If it's the latter, that is much less convenient than just setting environment variables.

Also what environment variables are now automatically passed to containers? Is it `http_proxy` , `https_proxy` , and `no_proxy` ? If these are passed, why is a config.json needed?

If `config.json` (is it `/etc/docker/config.json` ?), can you give me an example?

BTW, I removed the `args` and rebuilt and it still did not work.

arsulegai: Ah! I assumed you were aware of these changes from your previous comment. Let me brief it now, `config.json` is a file to be created on the client machine where we want to run the containers. Create this file under `~/.docker/` directory. This file will have the `****_proxy` and other env variables which are to be passed by the docker during build and running the container.

The `config.json` file looks like
```
{
    "proxies":
    {
        "default":
        {
            "httpProxy": "http://proxy-server-domain:<port>",
            "httpsProxy": "http://proxy-server-domain:<port>",
            "noProxy": "*.my-company.com,localhost,rest-api*",
            "hkpProxy": "http://proxy-server-domain:<port>"
        }
    }
}
```

Please let me know if adding this file helped you

danintel:
I updated the RFC to use `~/.docker/config.json` after testing it and getting it working.
The tricky part was making sure the Docker internal networks were included in `noProxy`.

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,164 @@
...
+  use `curl` and `apt-key add -` . For example, instead of the following in a
+  `Dockerfile` (which no longer works):
+
+```
+RUN \
+ apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
+ --recv-keys 8AA7AF1F1091A5FD
+```
+
+Use the following, which works in proxy (with http_proxy) and non-proxy environments:
```
dcmiddle: strike the parenthetic, based on recommendation to use the config.json file.

danintel: Fixed.  Please review and approve.

vaporos:
I like where where this has gone--less proxy support within the docker files themselves, but specific information on how to setup the proxy configuration.

The majority of the RFC contains explanation on how to use docker, instead of changes we need to make with Sawtooth. The changes are to remove what we currently have for proxy support. Do we still want to move forward with the RFC (which would now be an RFC to not have explicit proxy support), or should we just move forward with PRs to simplify the files and document how to setup proxy (which could be derived from the text of this RFC).

danintel:
Yes, I want to move forward as a RFC. The one part where (minor) changes are still needed is for key servers. The rest is good for informational reasons.

Follow-up would be removing unneeded proxy configuration/passthrough in Docker files, as they are found.

Dan

dcmiddle:
> I like where where this has gone--less proxy support within the docker files themselves, but specific information on how to setup the proxy configuration.
> 
> The majority of the RFC contains explanation on how to use docker, instead of changes we need to make with Sawtooth. The changes are to remove what we currently have for proxy support. Do we still want to move forward with the RFC (which would now be an RFC to not have explicit proxy support), or should we just move forward with PRs to simplify the files and document how to setup proxy (which could be derived from the text of this RFC)?

That seems like a valid reason for an RFC.. here's the architectural pattern we've adopted across all of the components and repos. We can refer to it to avoid inadvertent bikeshedding. It may be a little lighter weight than some of the other RFCs, but all the better if we can move things along expediently.

danintel:
Can I get approvals here for this Proxy RFC now that the issues are resolved? I see one approval from Dan M.

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+calls. The containers need special configuration to be executed behind
+network proxies. There are multiple ways to provide that proxy configuration
+and it is desirable that the Sawtooth repositories standardize on a single
+approach. The challenge is further complicated by different abilities and
+constraints of different versions of Docker and Ubuntu.
+
+Starting with Ubuntu 18 (Bionic), `gpg` , invoked through `apt-key adv` ,
+now ignores `http_proxy` and `https_proxy` environment variables, and ignores
+the command line options
+`apt-key adv --keyserver-options http-proxy=<proxy server url>`
```
arsulegai: Ubuntu 18 doesn't ignore command line options, however it would be too much of a code in every Dockerfile if we choose this way.

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+
+None
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The current alternative used is hand-modifying Docker scripts each time a
+proxy is required. This is time consuming, expensive, and unnecessary.
+
+An earlier alternative of using `apt-key adv` and passing proxy variables on
```
arsulegai: This paragraph needs changes. Xenial did not require `--keyserver-options` explicitly passing proxy env. `config.json` in `~/.docker/` is sufficient for Xenial. The concern is however with Bionic, where all Sawtooth Dockerfiles are moved to.

text/0045-docker-proxy.md
danintel: The current version of `gpg` does ignore the `http-proxy` and `https-proxy` command line options. I found this out from testing on Bionic and confirmed it with multiple documents on the web. This document, for example, says:

```http-proxy=value
...
These options have no more function since GnuPG 2.1```

https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html

File: text/0045-docker-proxy.md (hunk 5 above)
Comment (danintel) [reply text/0045-docker-proxy.md]: I do not mention that `--keyserver-options` is required (one can also use environment variables, for example, in Xenial).. I do mention that `--keyserver-options` was used on line 36 (not lines 140-143).

It is good to know `config.json` is also supported in Xenial, as it makes this proposal to use `config.jason` stronger.

File: text/0045-docker-proxy.md (hunk 4 above)
Comment (arsulegai) [reply text/0045-docker-proxy.md]: Ah! I am not sure what version of gpg gets installed. These CLI worked for me. Thanks for pointing it out.

File: text/0045-docker-proxy.md [hunk 6]
```diff
@@ -0,0 +1,167 @@
...
+`/etc/apt/apt.conf.d/proxy` .
+
+
+# Goals
+[goals]: #goals
+
+* Have Docker and Docker Compose files and shell scripts that work in explicit
+  proxy, transparent proxy, and non-proxy environments
+* Have a uniform method to access proxy servers across Hyperledger Sawtooth
+  projects * Minimize Docker and script clutter with unnecessary tests and
```
rbuysse: Is "* Minimize Docker..." meant to be its own bullet point?

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+RUN \
+ apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
+ --recv-keys 8AA7AF1F1091A5FD
+```
+
+Use the following, which works in proxy and non-proxy environments:
+
+```
+RUN \
+ apt-get update && apt-get install -yq curl && curl -sSL \
```
rberg2: I would like to see `--retry 3` or some other number in the curl arguments since we sometimes see intermittent failures due to downloading keys from the Ubuntu keyserver. Currently we have a fallback to the p80 pool if this server fails. Thoughts?

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+An earlier alternative of using `apt-key adv` and passing proxy variables on
+the gpg command line (discussed above) worked in Ubuntu Xenial, but not in
+Bionic. This alternative has the disadvantage of making Docker scripts messy
+with a lot of proxy configuration setup.
+
+The alternative of setting `https_proxy` , `http_proxy` , and `no_proxy`
+variables has the disadvantage of requiring these variables to be explicitly
+passed through the `args:` and `environment:` sections for each container in a `docker-compose.yaml` file.
+
+A forth alternative, setting configuration for newer versions of gpg (in
```
rbuysse: s/forth/fourth

text/0045-docker-proxy.md
arsulegai: Sounds good. How about having a curl try on fallback to the p80 pool if first 3 attempts fail?

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+
+* Have Docker and Docker Compose files and shell scripts that work in explicit
+  proxy, transparent proxy, and non-proxy environments
+* Have a uniform method to access proxy servers across Hyperledger Sawtooth
+  projects * Minimize Docker and script clutter with unnecessary tests and
+  settings
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
```
rbuysse: I'd prefer that this link to the official docker documentation rather than try to duplicate it in case the process changes in the future.

rbuysse:
I am still skeptical that this is appropriate material for an RFC but I approve of the direction this is taking and I will not disagree if the consensus is to codify this here.

text/0045-docker-proxy.md
danintel: The original attempt was to avoid a lot of boilerplate code. If adding boilerplate code is now OK with everyone, I can do that.

text/0045-docker-proxy.md
danintel: Yes. Looks like a copy paste error from the original Google Doc discussion.

danintel: Fixed.

text/0045-docker-proxy.md
arsulegai: No comments

text/0045-docker-proxy.md
danintel: I added a link to the official Docker document. It does not cover the special case of key servers though.

text/0045-docker-proxy.md
danintel: I added `--retry 3` and the  `p80.pool.sks-keyservers.net` keyserver. Both of these are not related to proxies, but it is also good practice.

text/0045-docker-proxy.md
danintel: Fixed

danintel:
I addressed all comments above for the proxy RFC. Let me know if there are more issues or questions.

text/0045-docker-proxy.md
rbuysse: Can you move the docker documentation link to this section instead? It's relevant in this section, but not to the one below.

text/0045-docker-proxy.md
rbuysse: Please include the gnupg link above in the RFC.

text/0045-docker-proxy.md
```diff
@@ -0,0 +1,176 @@
...
+ apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
+ --recv-keys 8AA7AF1F1091A5FD
+```
+
+Use the following, which works in proxy and non-proxy environments.
+Although not proxy-related, this example also has retry logic and uses
+two keyservers for robustness:
+
+```
+RUN \
```
rbuysse: Since this RFC is going to become the reference document, I'd like to see this section be more concise and readable.

If we're adding retries, I think we can dispense with the backup keyserver.

Also, this doesn't work unless gnupg2 is installed.

Suggestion:
```
RUN apt-get update \
 && apt-get install -yq \
 curl \
 gnupg2

RUN curl --retry 3 -sSL \
    'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD' \
    | apt-key add -
```

rberg2: This is what I had in mind with my earlier comment about using a curl retry.

text/0045-docker-proxy.md
danintel: Yes.

text/0045-docker-proxy.md
danintel: I removed the alternate keyserver as suggested by others and kept the `--retry 3`
Retry logic and alternate keyservers are not really a proxy issue.

text/0045-docker-proxy.md
danintel: Added a link.

danintel:
I addressed new issues raised in the current round of comments.  So please take a look here.

rbuysse:
Can you double check my comments here? https://github.com/hyperledger/sawtooth-rfcs/pull/45#discussion_r307814343

Looks like you removed the backup keyserver but my other concerns weren't addressed.

text/0045-docker-proxy.md
danintel: Package gnupg2 is not needed and it works for me without it being installed.  I only see documentation in the package:

```
dpkg-query -L gnupg
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/gnupg
/usr/share/doc/gnupg/DETAILS.gz
/usr/share/doc/gnupg/FAQ
/usr/share/doc/gnupg/HACKING.gz
/usr/share/doc/gnupg/OpenPGP.gz
/usr/share/doc/gnupg/README.Debian
/usr/share/doc/gnupg/README.gz
/usr/share/doc/gnupg/copyright
/usr/share/info
/usr/share/info/gnupg-card-architecture.png
/usr/share/info/gnupg-module-overview.png
/usr/share/info/gnupg.info-1.gz
/usr/share/info/gnupg.info-2.gz
/usr/share/info/gnupg.info.gz
/usr/share/doc/gnupg/NEWS.Debian.gz
/usr/share/doc/gnupg/NEWS.gz
/usr/share/doc/gnupg/THANKS.gz
/usr/share/doc/gnupg/TODO
/usr/share/doc/gnupg/changelog.Debian.gz
```

danintel: After removing packages `gnupg` and documentation package `gnupg2`, both `apt-key adv` and `apt-key add` both still work.  See example below.

```
$ sudo apt remove gnupg gnupg2
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  gnupg gnupg2
0 upgraded, 0 newly installed, 2 to remove and 0 not upgraded.
After this operation, 423 kB disk space will be freed.
Do you want to continue? [Y/n] y
(Reading database ... 184249 files and directories currently installed.)
Removing gnupg2 (2.2.4-1ubuntu1.2) ...
Removing gnupg (2.2.4-1ubuntu1.2) ...
Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
$ sudo apt-key del  44FC67F19B2466EA                              
OK
$ apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 44FC67F19B2466EA
Executing: /tmp/dano/apt-key-gpghome.yCD8U38W9A/gpg.1.sh --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 44FC67F19B2466EA
gpg: key 44FC67F19B2466EA: public key "Hyperledger Sawtooth Nightly Signing Key <distributedledger@intel.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: no writable keyring found: Not found
gpg: error reading '[stdin]': General error
gpg: import from '[stdin]' failed: General error
gpg: Total number processed: 0
$ sudo apt-key del  44FC67F19B2466EA                              
OK
$ curl --retry 3 -sSL 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x44FC67F19B2466EA' | sudo apt-key add -                
OK
```

danintel:
@rbysse I checked all your comments and made some changes.  If there are any more issues, please let me know which ones.  Here is a list:

* _Please include the gnupg link above in the RFC._
Added a link around line 42.

* _I'd prefer that this link to the official docker documentation rather than try to duplicate it in case the process changes in the future._
I deleted the text and replaced it with a link under "Guide-level explanation"
around line 62.

* _Can you move the docker documentation link to this section instead? It's relevant in this section, but not to the one below._
Same as above.

* _Is "* Minimize Docker..." meant to be its own bullet point?_
Yes.  Now a separate line on line 52.

* _s/forth/fourth_
Fixed at line 124.
```

* _I am still skeptical that this is appropriate material for an RFC but I approve of the direction this is taking and I will not disagree if the consensus is to codify this here._
OK. The docker-compose/apt-key adv/gpg interaction isn't really documented properly.

*_Since this RFC is going to become the reference document, I'd like to see this section be more concise and readable.  If we're adding retries, I think we can dispense with the backup keyserver. RUN apt-get update && apt-get install -yq curl gnupg2
RUN curl --retry 3 -sSL 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD' | apt-key add -_

I added the backup keyserver from an earlier suggestion. I removed it again and now there is just the `--retry`

*_Also, this doesn't work unless gnupg2 is installed._
Package gnupg2 is just documentation under /usr/share/doc and /usr/share/info
See a previous comment for gnupg2 package contents.  Perhaps you mean gnupg?  But, I removed both gnupg and gnupg2 and was able to install keys with `apt-key add` and `apt-key add`.

**Comment by rbuysse** [comment]
I'm not sure how you're testing this, but the example code you have in the RFC doesn't work in docker.

Trying to build the following dockerfile:

```
FROM ubuntu:bionic

RUN \
 apt-get update && apt-get install -yq curl && curl --retry 3 -sSL \
    'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD' \
    | apt-key add -
```

Exits with error:
```
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
(23) Failed writing body
The command '/bin/sh -c apt-get update && apt-get install -yq curl && curl --retry 3 -sSL     'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD'     | apt-key add -' returned a non-zero code: 255
```

The suggestion I provided is more readable, breaking up the two operations into logical steps and it works.

**Comment by danintel** [comment]
@rbuysse I have no problem running Sawtooth and apt-get and apt-key without explicitly installing gnupg/gnupg2/gnupg1.  I just tried it on Ubuntu 18.04 LTS, docker-compose 1.24.1.

I also notice that none of the sawtooth-core (and no sawtooth-* repo I checked) ever installs gnupg2 or gnupg1.  Repo sawtooth-core installs gnupg, but that is documentation only, no binary. 

So if package gnupg2 (or gnupg1) is required, why is no Sawtooth Dockerfile installing it?

Also, installing a key is required in both proxy and non-proxy environments and properly should be documented separately. I also notice that the Sawtooth documentation does not mention installing gnupg2:
https://sawtooth.hyperledger.org/docs/core/releases/latest/sysadmin_guide/installation.html

But there is something different about your environment from all others, so I am adding package `gnupg2` to the list. However, this (apt-key dependencies) is really a non-proxy issue.

I updated the RFC so please have a look and approve it. Thanks

**Comment by peterschwarz** [comment]
What happens when your run a Docker file containing just this:

```Dockerfile
FROM ubuntu:bionic

RUN \
 apt-get update && apt-get install -yq curl && curl --retry 3 -sSL \
    'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8AA7AF1F1091A5FD' \
    | apt-key add -
```
as @rbuysse suggested?

rberg2:
Hi Dan, Gnupg does need to installed for secure apt to work or apt won't be able to verify the signed package. Since Ubuntu Bionic gnupg is not installed by default in the Docker images. Gnupg was added to sawtooth-core [here](https://github.com/hyperledger/sawtooth-core/commit/c23f330a2520e625b58cbeda446f230fca41246e). The gnupg package depends on gnupg-utils and gpg 2 among others and installing it results in a complete installation of gnupg.

danintel:
@peterschwarz  It fails, but works when I add `gnupg2`
My guess is the other Docker instances are using pre-built images with gnupg2 previously installed.

In any case, I already updated the docs to specify `gnupg2`

The existing Sawtooth documentation should also be updated to mention that  one of `gnupg2/gnupg1/gnupg` is required. That is not done anywhere.

Again this is not a proxy issue and I already updated this RFC.

danintel:
@rberg2 The PR has no documention updates. So the Sawtooth user/administrator is never told to install gnupg2 (or gnupg or gnupg1).

In any case this RFC mentions gnupg2 and is now ready to be approved.

danintel:
I created a separate PR in sawtooth-core to add missing documentation about installing `gnupg2` when installing Sawtooth. Please see https://github.com/hyperledger/sawtooth-core/pull/2222

rbuysse:
Also I'd still like to see the docker example code cleaned up as mentioned in this comment: https://github.com/hyperledger/sawtooth-rfcs/pull/45#discussion_r307814343

danintel:
@rbuysse Reading between the lines, I assume the important part of your comment is having two `RUN` stanzas, not one.  I updated the RFC to have two `RUN` stanzas.

danintel:
@rberg2 I removed the comment about two keyservers.

vaporos:
We should either be clear that this applies only to the specific Ubuntu versions, or we should make it more principle-based so that it would apply to CentOS/RHEL as well.

danintel:
@vaporos Ubuntu is specifically mentioned under "Summary". It does not apply to CentOS/RHEL as they do not use APT, at least in general.

I could emphasize the restriction more.

PR #46, opened 2019-06-22
Update Swift SDK RFC with PR number

agunde406:
adds PR number to the rfc and changes the file name to include PR number


agunde406:
@danintel That change is already made in the second commit

PR #47, opened 2019-07-01
Fix Swift SDK RFC 0037 Numbering

danintel:
The Swift SDK RFC is 0037, not 0000.

Signed-off-by: danintel <daniel.anderson@intel.com>


agunde406:
This is a duplicate of #46

danintel:
Looks like PR #46 https://github.com/hyperledger/sawtooth-rfcs/pull/46 incorporated the file renaming/correction change, so this is no longer needed.

